PR #6, opened 2022-11-17
CuPy support

### LINKED ISSUE #3: Support `cupy.ndarray`?
Create an issue and assign to self so I don't forget. cc: @kmaehashi @asi1024 @emcastillo

asmeurer:
This refactors the code to be array library independent, so that it can work with either NumPy or CuPy.

Other libraries could be supported as well, although the wrapping for each library would be specific to the differences between that library and the array API. 

Fixes #3.


numpy_array_api_compat/_aliases.py
```diff
@@ -158,148 +207,192 @@ def asarray(
...
     device: Optional[Device] = None,
-    copy: Optional[Union[bool, np._CopyMode]] = None,
+    copy: "Optional[Union[bool, np._CopyMode]]" = None,
+    namespace = None,
 ) -> ndarray:
+    if namespace is None:
+        try:
+            xp = get_namespace(obj, _use_compat=False)
+        except ValueError:
+            # TODO: What about lists of arrays?
```
rgommers, line 217: What you have seems reasonable. Libraries should be passing `namespace=numpy` if they want this to work, or otherwise handle lists/sequences themselves.

rgommers:
I'm looking at `get_xp` and thinking there's still a piece of the puzzle missing here. Should I wait until that's in place, or am I just missing how it works?

asmeurer:
I haven't added any cupy functionality yet if that's what you mean. I also need to make sure that NumPy is only an optional dependency of all the code here.

asmeurer:
This is ready to go. Anyone want to review the overall design here? I've described it in the README.

I still need to set up some CI here, but I've tested the NumPy and CuPy support with the array API test suite and the only issues are things which can't really be fixed by wrapping.

array_api_compat/common/_helpers.py
```diff
@@ -0,0 +1,175 @@
...
+    return isinstance(x, (np.ndarray, np.generic))
+
+def _is_cupy_array(x):
+    # Avoid importing NumPy if it isn't already
+    if 'cupy' not in sys.modules:
+        return False
+
+    import cupy as cp
+
+    # TODO: Should we reject ndarray subclasses?
```
rgommers, line 29: I am not aware of "bad" `cupy.ndarray` subclasses that need to be avoided.

array_api_compat/common/_helpers.py
```diff
@@ -0,0 +1,175 @@
...
+import sys
+
+def _is_numpy_array(x):
+    # Avoid importing NumPy if it isn't already
+    if 'numpy' not in sys.modules:
+        return False
+
+    import numpy as np
+
+    # TODO: Should we reject ndarray subclasses?
```
rgommers, line 19: Always a question mark. With "reject", do you mean "convert to ndarray", or "raise", or "ignore", or ....?

rgommers (approved):
I like the approach Aaron, it's pretty easy to understand and it should be extensible. 

The one thing that comes to mind as a possible concern is that the `get_xp` decorator may have some overhead that's larger than desired - if so that could be fairly easily changed to a codegen approach where `xp` gets replaced by `np` or `cp` it looks like.

asmeurer:
Here's an idea of the overhead of get_xp:

```py
In [1]: import numpy as np

In [2]: import array_api_compat.numpy as npc

In [3]: a = np.asarray(1.)

In [4]: %timeit np.arccos(a)
435 ns ± 8.46 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [5]: %timeit npc.acos(a)
737 ns ± 4.31 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

It looks like it's about 300 ns per function call. Does this seem acceptable or should we investigate faster approaches?

array_api_compat/common/_helpers.py
asmeurer: My thinking here was specifically for `get_namespace`, although this helper is also used in `asarray`. Right now it does accept them, because it uses `isinstance`. I've no idea if an ndarray subclass could implement behavior that would break things here. Presumably if they are properly Liskov substitutable, they shouldn't. My guess is that at worst some functions might return ndarray instead of the appropriate subclass, because we don't properly account for subclases when creating new array objects.

But I haven't tested it. It's possible some people one of the units libraries (I forget which ones subclass ndarray) might care about this at some point, though.

rgommers: > Presumably if they are properly Liskov substitutable, they shouldn't.

`np.matrix` for one isn't, that's a major issue. IIRC also `np.ma.MaskedArray` isn't.

PR #14, opened 2023-01-09
WIP: PyTorch compatibility layer

asmeurer:


array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    # ints and uints (mixed sign)
+    (torch.int8, torch.uint8): torch.int16,
+    (torch.int16, torch.uint8): torch.int16,
+    (torch.int32, torch.uint8): torch.int32,
+    (torch.int64, torch.uint8): torch.int64,
+    (torch.uint8, torch.int8): torch.int16,
+    (torch.uint8, torch.int16): torch.int16,
+    (torch.uint8, torch.int32): torch.int32,
+    (torch.uint8, torch.int64): torch.int64,
+    # floats
```
lezcano, line 59: No complex for now, I guess?

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    return torch.amin(x, axis, keepdims=keepdims)
+
+# torch.sort also returns a tuple
+# https://github.com/pytorch/pytorch/issues/70921
+def sort(x: array, /, *, axis: int = -1, descending: bool = False, stable: bool = True, **kwargs) -> array:
+    return torch.sort(x, dim=axis, descending=descending, stable=stable, **kwargs).values
+
+def _normalize_axes(axis, ndim):
+    axes = []
+    lower, upper = -ndim, ndim - 1
```
lezcano, line 175: You may have issues here when the input vector is a 0-dim, as the range you get is empty.

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    for a in axis:
+        if a < lower or a > upper:
+            # Match torch error message (e.g., from sum())
+            raise IndexError(f"Dimension out of range (expected to be in range of [{lower}, {upper}], but got {a}")
```
lezcano, line 179: Nit. You may want to have a utility like this one for readability
https://github.com/pytorch/pytorch/blob/f3bf46e801dec2637751224fd6e27fbf97453bc6/torch/_prims_common/__init__.py#L1556-L1566
then you could do
```python
check(lower <= a <= upper, lambda: f"...")
```
Note that the use of the lambda is just to avoid doing all the string parsing unconditionally.

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+        if a < 0:
+            a = a + ndim
+        if a in axes:
+            # Match torch error message but use IndexError instead of RuntimeError
+            raise IndexError(f"dim {a} appears multiple times in the list of dims")
```
lezcano, line 184: Do we call them `dims`, `axes` or...?

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+def _normalize_axes(axis, ndim):
+    axes = []
+    lower, upper = -ndim, ndim - 1
+    for a in axis:
+        if a < lower or a > upper:
+            # Match torch error message (e.g., from sum())
+            raise IndexError(f"Dimension out of range (expected to be in range of [{lower}, {upper}], but got {a}")
+        if a < 0:
+            a = a + ndim
+        if a in axes:
```
lezcano, line 182: Note. The current algorithm is quadratic. Since we are doing a `sort` at the end, we may want to leverage that to drop the quadratic dependence on the data.
Now, `axes` will, in general, be short, so if you think this is a non-issue, I'm fine with it as well.

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+            # Match torch error message but use IndexError instead of RuntimeError
+            raise IndexError(f"dim {a} appears multiple times in the list of dims")
+        axes.append(a)
+    return sorted(axes)
+
+
+def _apply_keepdims(x, ndim, keepdims):
+    if keepdims:
+        return x[(None,)*ndim]
```
lezcano, line 191: This seems wrong. You are not passing the sizes of the of the initial vector, so how do you know what's the shape the tensor `x` needs to have. This should be done in PyTorch via `unsqueeze`s and a `expand`. See for example:
https://github.com/pytorch/pytorch/blob/f3bf46e801dec2637751224fd6e27fbf97453bc6/torch/_refs/__init__.py#L2057-L2060
and
https://github.com/pytorch/pytorch/blob/f3bf46e801dec2637751224fd6e27fbf97453bc6/torch/_prims/__init__.py#L1246-L1256

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    x = torch.asarray(x)
+    ndim = x.ndim
+
+    # https://github.com/pytorch/pytorch/issues/29137. Separate from the logic
+    # below because it still needs to upcast.
+    if axis == ():
+        if dtype is None:
+            if x.dtype in [torch.int8, torch.int16, torch.int32]:
+                return x.to(torch.int64)
+            # we can't upcast uint8 because there is no torch.uint64
```
lezcano, line 210: There is no `uint64`, but you can certainly fit a `uint8` into a `int64`, so what's the issue here?

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    x = torch.asarray(x)
+    ndim = x.ndim
+
+    # https://github.com/pytorch/pytorch/issues/29137.
+    # Make sure it upcasts.
+    if axis == ():
+        if dtype is None:
+            if x.dtype in [torch.int8, torch.int16, torch.int32]:
+                return x.to(torch.int64)
+            # we can't upcast uint8 because there is no torch.uint64
```
lezcano, line 251: Same as above.

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+
+# torch.concat doesn't support dim=None
+# https://github.com/pytorch/pytorch/issues/70925
+def concat(arrays: Union[Tuple[array, ...], List[array]],
+           /,
+           *,
+           axis: Optional[int] = 0,
+           **kwargs) -> array:
+    if axis is None:
+        arrays = tuple(ar.ravel() for ar in arrays)
```
lezcano, line 405: They PyTorch name for `ravel` is `flatten`, which I think is nicer. I'm fine with either though.

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+    for a in sequence:
+        x = torch.squeeze(x, a)
+    return x
+
+# The axis parameter doesn't work for flip() and roll()
+# https://github.com/pytorch/pytorch/issues/71210. Also torch.flip() doesn't
+# accept axis=None
+def flip(x: array, /, *, axis: Optional[Union[int, Tuple[int, ...]]] = None) -> array:
+    if axis is None:
+        axis = tuple(range(x.ndim-1, -1, -1))
```
lezcano, line 430: Why reversed?

array_api_compat/torch/_aliases.py
```diff
@@ -0,0 +1,560 @@
...
+         dtype: Optional[Dtype] = None,
+         device: Optional[Device] = None,
+         **kwargs) -> array:
+    if isinstance(shape, int):
+        shape = (shape,)
+
+    return torch.full(shape, fill_value, dtype=dtype, device=device, **kwargs)
+
+# Functions that aren't in torch https://github.com/pytorch/pytorch/issues/58742
+def expand_dims(x: array, /, *, axis: int = 0) -> array:
```
lezcano, line 513: Odd that this function that's called `dims` in plural just accepts an integer. We also have a small dissonance again here between `axes` and `dims`, but well.

lezcano (commented):
This looks quite good. I left a number of comments, mostly with suggestions for improvements. In general, it'd be good to add tests (and linting!) for the issues that this review may have found as well.

Now, why are there some functions that preprocess their inputs with `torch.asarray` and others that don't?

A comment on reductions. All the reductions that don't support multiple axes may be `torch.movedim`, moving all the relevant dimensions to the end, then a call to `flatten` and a call to the reduction with `axis=-1`. This would incur in one copy and one call to the kernel, so it should be *much* more efficient in eager mode, and easier to optimise for the compiler. It should be easy to do this in a generic way via a wrapper.

array_api_compat/common/_helpers.py
```diff
@@ -29,11 +30,24 @@ def _is_cupy_array(x):
     # TODO: Should we reject ndarray subclasses?
     return isinstance(x, (cp.ndarray, cp.generic))
 
+def _is_torch_array(x):
+    # Avoid importing torch if it isn't already
+    if 'torch' not in sys.modules:
+        return False
```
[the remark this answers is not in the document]
asmeurer: I don't want to import the module if it hasn't been imported already.

array_api_compat/torch/_aliases.py
asmeurer: In this case I just matched the torch error message. But I can use axes to match the spec if we want.

array_api_compat/torch/_aliases.py
asmeurer: This was just taken from NumPy. I know we discussed this for some of the functions but I can't find any discussion for expand_dims specifically.

array_api_compat/torch/_aliases.py
asmeurer: Yes, I want to just focus on the 2022.12 version of the spec for now. We can add 2023.12 support later. Ideally I'd like to have test suite support first. There's a couple of things that were added in 2023 in addition to complex numbers.

array_api_compat/common/_helpers.py
```diff
@@ -169,7 +188,16 @@ def to_device(x: "Array", device: "Device", /, *, stream: Optional[Union[int, An
     elif _is_cupy_array(x):
         # cupy does not yet have to_device
         return _cupy_to_device(x, device, stream=stream)
-
+    elif _is_torch_array(x):
+        return _torch_to_device(x)
```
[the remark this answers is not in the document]
asmeurer: I use pyflakes in my editor but it doesn't handle this kind of error. I checked `pylint -E` and this is the only error it found. There's also some problems with annotations, but I've honestly been just ignoring that for now until someone complains about it.

asmeurer:
> This looks quite good. I left a number of comments, mostly with suggestions for improvements. In general, it'd be good to add tests (and linting!) for the issues that this review may have found as well.

Yes, I want to add the test suite to the CI. I will finish #13 before releasing this. I guess we can add some basic linting, although we'll need some ignore comments due to the import * stuff that I have to do in the init files. 

> Now, why are there some functions that preprocess their inputs with torch.asarray and others that don't?

I did that in the functions that I needed to access attributes on the input (`x.ndim`). Otherwise I just pass everything through to the corresponding torch functions. It's quite possible I missed some cases here as the test suite is only going to be testing inputs that are valid according to the spec (i.e., only passing array inputs to functions).  

> A comment on reductions. All the reductions that don't support multiple axes may be implemented via a call to torch.movedim, moving all the relevant dimensions to the end, then a call to flatten and a call to the reduction with axis=-1. This would incur in one copy and one call to the kernel, so it should be much more efficient in eager mode, and much easier to optimise for the compiler. It should be easy to do this in a generic way via a wrapper.

I'll take a look.

array_api_compat/common/_helpers.py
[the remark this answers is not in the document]
lezcano: About annotations: Sure, makes sense.
Also, good to know this is the only omision in the file :)

array_api_compat/torch/_aliases.py
lezcano: whatever the specs settled for

array_api_compat/torch/_aliases.py
asmeurer: f-strings aren't evaluated unless they are executed.

array_api_compat/torch/_aliases.py
asmeurer: Maybe I should just copy the similar NumPy helper https://github.com/numpy/numpy/blob/be52f19259d47cd27095e186a7974710c60169ee/numpy/core/numeric.py#L1328, which compares sets. Or if pytorch has an internal helper along these lines that I can just import that would be even better.

array_api_compat/torch/_aliases.py
asmeurer: Just need to rename the helper here. I only need to manually apply keepdims when `axis=None`, because of https://github.com/pytorch/pytorch/issues/71209. So it's really just a special case of expanding a 0-D tensor to a (1, 1, ..., 1) tensor. If there's a better way to do it than `newaxis` indexing I can do that instead.

array_api_compat/torch/_aliases.py
lezcano: In PyTorch we have what I believe it's a slightly less general method:
https://github.com/pytorch/pytorch/blob/e61d5b95886c11340118f559529397a7d0556d6b/torch/_prims_common/__init__.py#L489
Feel free to copy NumPy's implementation if you believe it's the better one

array_api_compat/torch/_aliases.py
lezcano: Renaming and a comment would be good, yep.
And yes, this is best done in PyTorch via a number of `unsqueeze`s

array_api_compat/torch/_aliases.py
lezcano: Actually note that in this case the f-string is being passed to the function `check` and thus it'd be evaluated there.

array_api_compat/torch/_aliases.py
asmeurer: The downside of the NumPy implementation (comparing sets) is that we can't include the repeated axis in the error message.

asmeurer: Is it OK to import and use this internal pytorch function here?

lezcano: May be best to copy it and leave a link for where it was copied from. Those internal functions may or may not exist in the future.

array_api_compat/torch/_aliases.py
asmeurer: I noticed this internal function in pytorch that does exactly this, although it isn't using either of these methods https://github.com/pytorch/pytorch/blob/e61d5b95886c11340118f559529397a7d0556d6b/torch/_prims/__init__.py#L1411.

lezcano: It does use `unsqueeze` and `expand` at the end, as it calls into `broadcast_in_dim`:
https://github.com/pytorch/pytorch/blob/f3bf46e801dec2637751224fd6e27fbf97453bc6/torch/_prims/__init__.py#L1246-L1256

asmeurer:
This is ready for final review. I have addressed all the review comments. I have also added CI with XFAILs/SKIPs (some tests have to be skipped because they crash on CI), and have updated the README with a list of the known issues. Assuming everything looks good here, I can cut a release with this.

lezcano:
how come this one was merged before the final review?

asmeurer:
Because I am on PTO this week and I wanted to get this merged and released before then. If you have any further comments feel free to comment here or open a new issue.

array_api_compat/torch/__init__.py
```diff
@@ -0,0 +1,19 @@
+from torch import *
+
+# Several names are not included in the above import *
```
albanD, line 3: Ho that is bad. I guess this is because we're missing things in the `__all__` attribute? If you have a list for these, could you open an issue on the pytorch repo for us to fix them?

albanD: Ho I forgot about that issue. Perfect! Thanks!

PR #20, opened 2023-03-03
Enables get_namespace to return namespace when input adopts Array API spec

thomasjpfan:
This PR enables `get_namespace` to return the Array API namespace if the input already adapts the Array API standard.

### DIFF
--- a/tests/test_get_namespace.py
+++ b/tests/test_get_namespace.py
@@ -12,3 +12,10 @@ def test_get_namespace(library):
     expected_namespace = getattr(array_api_compat, library)
     assert namespace is expected_namespace
 
+
+@pytest.mark.parametrize("array_namespace", ["cupy.array_api", "numpy.array_api"])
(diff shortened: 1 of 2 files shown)


kgryte (approved):
LGTM. This looks to be a bug in the current implementation.

tests/test_get_namespace.py
```diff
@@ -12,3 +12,10 @@ def test_get_namespace(library):
     expected_namespace = getattr(array_api_compat, library)
     assert namespace is expected_namespace
 
+
+@pytest.mark.parametrize("array_namespace", ["cupy.array_api", "numpy.array_api"])
+def test_get_namespace_returns_actual_namespace(array_namespace):
```
asmeurer, line 17: Why is this a separate test from the one above?

asmeurer: Going to just merge this and fix this and the other issues you opened issues for in another PR.

PR #25, opened 2023-03-06
Various fixes for the torch wrapper

asmeurer:
Primarily from @thomasjpfan's feedback.


asmeurer:
This is ready for review.

rgommers (approved):
These changes LGTM. One CI job for numpy 1.21 still has 8 failures, which all don't look too interesting (~1e-46 inputs to some ufuncs).

asmeurer:
No worried on CI. It typically takes a few iterations to get all the possible XFAILs from hypothesis.

asmeurer:
(and just to be clear, those failures are all because of value-based type promotion, which we cannot disable in NumPy 1.21)

asmeurer:
Now NumPy is crashing on CI with a floating-point exception https://github.com/asmeurer/array-api-compat/actions/runs/4368929147/jobs/7642133923. There must be some issue with mod in the C math library used on GitHub Actions.

PR #29, opened 2023-03-08
Rename get_namespace to array_namespace

### LINKED ISSUE #19: rename or alias `get_namespace` to `array_namespace`
Having a function `array_namespace` seems more intuitive than `get_namespace`, mainly due to symmetry with e.g. `repr()`, `str()`, `iter()`, `next()`, since `array_namespace(x)` returns `x.__array_namespace__()` if it exists, and does something sensible if it does not.

asmeurer:
get_namespace is kept as a backwards compatible wrapper to array_namespace

Making this as a separate PR from #25 as it might be more controversial.

Fixes #19.


asmeurer:
(this is built on #25. The diff should clear once that is merged)

asmeurer:
Thoughts on this? I want to release the fixes from #25 but if we are going to do this we should do it now.

rgommers:
It seems fine to me. Honestly I'm not sure it's that much better a name, but `get_namespace` is also not a name we've really committed to anywhere. IIRC @shoyer came up with this name for NEP 47, so let's see if he minds this change.

shoyer:
Sounds good to me, too. I don't recall putting much thought into this name.

PR #32, opened 2023-03-14
Implement isdtype()

### LINKED ISSUE #27: Add Data Type Function, `isdtype`, from the v2022 spec
Add Data Type Function, `isdtype`, from the [v2022 spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.isdtype.html#array_api.isdtype).

asmeurer:
Fixes #27 

I've added some rudimentary testing here in lieu of test suite support, and also testing that the implementation works with some non-spec dtypes.

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from functools import wraps
-from builtins import all as builtin_all
+from builtins import all as builtin_all, any as builtin_any
 
 from ..common._aliases import (UniqueAllResult, UniqueCountsResult,
                                UniqueInverseResult,
@@ -18,13 +18,17 @@ if TYPE_CHECKING:
 import torch
 array = torch.Tensor
 
-_array_api_dtypes = {
-    torch.bool,
+_int_dtypes = {
     torch.uint8,
     torch.int8,
     torch.int16,
     torch.int32,
     torch.int64,
+}
+
+_array_api_dtypes = {
+    torch.bool,
+    *_int_dtypes,
     torch.float32,
     torch.float64,
 }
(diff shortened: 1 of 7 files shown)


array_api_compat/torch/_aliases.py
```diff
@@ -602,6 +606,43 @@ def tensordot(x1: array, x2: array, /, *, axes: Union[int, Tuple[Sequence[int],
...
+    https://data-apis.org/array-api/latest/API_specification/generated/array_api.isdtype.html
+    for more details
+    """
+    if isinstance(kind, tuple) and _tuple:
+        return builtin_any(isdtype(dtype, k, _tuple=False) for k in kind)
+    elif isinstance(kind, str):
+        if kind == 'bool':
+            return dtype == torch.bool
+        elif kind == 'signed integer':
+            return dtype in _int_dtypes and dtype.is_signed
```
asmeurer, line 630: Is there a better way to test if a torch dtype is integral.

array_api_compat/torch/_aliases.py
```diff
@@ -602,6 +606,43 @@ def tensordot(x1: array, x2: array, /, *, axes: Union[int, Tuple[Sequence[int],
...
+        elif kind == 'unsigned integer':
+            return dtype in _int_dtypes and not dtype.is_signed
+        elif kind == 'integral':
+            return dtype in _int_dtypes
+        elif kind == 'real floating':
+            return dtype.is_floating_point
```
asmeurer, line 636: Is this the correct way to check for torch dtype categories?

array_api_compat/torch/_aliases.py
rgommers: There isn't. Based on dtype properties it would be:
```python
not dtype.is_floating_point and not dtype.is_complex and dtype.is_signed and dtype != dtype.bool
```
which looks worse and may also not be foolproof. The check you have will need explicit updating if for example `uint16` is added to PyTorch, but that's okay (nothing like that is in the pipeline AFAIK).

array_api_compat/torch/_aliases.py
rgommers: This should be good, yes.

array_api_compat/torch/_aliases.py
asmeurer: Hopefully this function itself would be added to pytorch by then.

PR #38, opened 2023-04-18
2022.12 support

asmeurer:
This is still WIP because we need some fixes to the test suite.


rgommers:
@asmeurer are you also adding `fft` extension support, in this PR or a separate one?

asmeurer:
The test suite doesn't check anything yet, but assuming `torch.fft` and so on already match the spec then there's nothing to do.

rgommers:
`unique`

> `unique_*` are harder to work around directly so I probably won't attempt that.

That's probably caused by [torch.unique](https://pytorch.org/docs/stable/generated/torch.unique.html) sorting by default and having a keyword for it; sorting won't work with complex dtypes. Not too relevant for the standard, but a pain to get to work indeed in a wrapper. So I agree with your decisions here.

honno (approved):
LGTM!

Regarding `array_api_compat.torch.take()`, IIRC `index_put()` wasn't quite adequate for "full" NumPy compat and I ended up using `torch.Tensor.__get__()` (https://github.com/Quansight-Labs/numpy_pytorch_interop/pull/96), but it's not that surprising if it's fine for the Array API scope. Note my `test_take` does have a few limitations...

https://github.com/data-apis/array-api-tests/blob/f82c7bc8627cc2c3a44fa3e425f53a253a609aa8/array_api_tests/test_indexing_functions.py#L21-L24

asmeurer:
We figured out @tupui's isdtype issue. It was related to a version of the function that he had implemented in scipy. It seems there aren't any issues with the version in the compat library. So we can go ahead and do a release here.

PR #40, opened 2023-04-28
ENH: support CuPy to_device "cpu"

tylerjereddy:
* allow `to_device(x, "cpu")` for CuPy--this makes it easier to write portable unit tests where the
expected value is on the host (a NumPy array)

* otherwise, you can end up writing different shims in downstream libraries to move the array-like back to the host (`.get()` for CuPy...); if CuPy conforms to the standard, we shouldn't need this shim long-term

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -154,6 +154,11 @@ def _cupy_to_device(x, device, /, stream=None):
 
     if device == x.device:
         return x
+    elif device == "cpu":
+        # allowing us to use `to_device(x, "cpu")`
+        # is useful for portable test swapping between
+        # host and device backends
+        return x.get()
     elif not isinstance(device, _Device):
         raise ValueError(f"Unsupported device {device!r}")
     else:
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,0 +1,23 @@
+from ._helpers import import_
+from array_api_compat import to_device, device
+
+import pytest
+import numpy as np
+from numpy.testing import assert_allclose
+
+@pytest.mark.parametrize("library", ["cupy", "numpy", "torch"])
+def test_to_device_host(library):
+    # different libraries have different semantics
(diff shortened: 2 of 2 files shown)


asmeurer:
Looks fine to me, although this might also be something to request upstream from cupy.

asmeurer:
> if CuPy conforms to the standard, we shouldn't need this shim long-term

The standard doesn't say anything about how devices are specified https://data-apis.org/array-api/latest/design_topics/device_support.html#syntax-for-device-assignment. The "cpu" device for NumPy was just a convention that was chosen, probably based off pytorch. Maybe the standard should specify that "cpu" == host? It could be worth opening an issue about. I'm not sure what has already been discussed here.

asmeurer:
CuPy won't be tested until I test it manually. I'll be sure to do that before making another release.

tylerjereddy:
For the record I did test on CuPy (and Torch) locally with a 1080 Ti and passed for me.

asmeurer:
I'm actually a little unsure about this thinking about it. If you transfer a CuPy array to NumPy, you change what namespace it uses. That's different from PyTorch where CPU and GPU tensors both use the `torch` namespace.

tylerjereddy:
Good point, I shared that in the cross-linked array API standard issue. That said, at least for my use case, I didn't care so much about staying in the same array namespace as being able to have the memory usable directly in `assert_allclose()`. The fact that different array libraries have such drastically different semantics in moving data between device and host probably further underlines the need for a portable shim of some sort? Unless you really want each lib to write its own shims for this.

asmeurer:
Well you would need to make sure to recall array_namespace after calling to_device. 

My feeling is that this goes against some of the design decisions of the standard (see https://data-apis.org/array-api/latest/purpose_and_scope.html and https://data-apis.org/array-api/latest/design_topics/data_interchange.html). But device stuff isn't really my wheelhouse so hopefully other folks will weigh in on the issue or at the consortium meeting.

PR #41, opened 2023-04-28
Use the main array-api-tests repo instead of my fork on CI

asmeurer:


asmeurer:
There's a few legitimate test failures from torch due to some updates in the test suite.

PR #47, opened 2023-07-15
Fix the torch.take() wrapper to make axis optional for ndim = 1

### LINKED ISSUE #34: Implement torch.take with `axis` argument
I suspect take's `axis` argument will be needed at some point. Can we add a simple implementation for PyTorch?

```python
def take(array, indices, *, axis):
    key = [slice(None)] * array.ndim
    key[axis] = indices
    return array[key]
```

asmeurer:
Closes #34

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -681,7 +681,11 @@ def isdtype(
     else:
         return dtype == kind
 
-def take(x: array, indices: array, /, *, axis: int, **kwargs) -> array:
+def take(x: array, indices: array, /, *, axis: Optional[int] = None, **kwargs) -> array:
+    if axis is None:
+        if x.ndim != 1:
+            raise ValueError("axis must be specified when ndim > 1")
+        axis = 0
     return torch.index_select(x, axis, indices, **kwargs)
 
 __all__ = ['result_type', 'can_cast', 'permute_dims', 'bitwise_invert', 'add',


array_api_compat/torch/_aliases.py
```diff
@@ -681,7 +681,11 @@ def isdtype(
     else:
         return dtype == kind
 
-def take(x: array, indices: array, /, *, axis: int, **kwargs) -> array:
+def take(x: array, indices: array, /, *, axis: Optional[int] = None, **kwargs) -> array:
+    if axis is None:
+        if x.ndim != 1:
+            raise ValueError("axis must be specified when ndim > 1")
+        axis = 0
```
lezcano, line 688: Or you can dispatch to `torch.take`, either should be alright.

lezcano (approved):

array_api_compat/torch/_aliases.py
asmeurer: torch.take doesn't support an axis argument.

lezcano: exactly :D `torch.take() == numpy.take(axis=None)`

lezcano: it's a bit more general, as it sees the whole tensor as a 0dim tensor and indexes into it, so it also works for `ndim > 1`.

asmeurer: Right. The flattening behavior is also there in NumPy, but I think we wanted to avoid that with the array API `take`. I'm not sure to what degree we should try to avoid that for the compat library, though (c.f. https://github.com/data-apis/array-api-compat/issues/34#issuecomment-1636813887)

PR #51, opened 2023-07-16
CI: Add publishing to PyPI and TestPyPI with trusted publishers

### LINKED ISSUE #50: Interest in publishing workflows?
At the moment it seems like there is no CI/CD publishing infrastructure for `array-api-compat`. Is there interest in setting up a GitHub Actions based workflow that uses [PyPI trusted publishers](https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/) to deploy to TestPyPI and PyPI?

The [workflow that we use for `pyhf`](https://github.com/scikit-hep/pyhf/blob/ff9cb94025e5485b23ea81a06ce8916055297c7f/.github/workflows/publish-package.yml) does this and publishes on GitHub Release or workflow dispatch (but can be easily configured to work with tags or anything).

matthewfeickert:
Resolves #50 

Use the OpenID Connect (OIDC) standard to publish to PyPI and TestPyPI using PyPI's "Trusted Publisher" implementation to publish without using API tokens stored as GitHub Actions secrets. Use an optional GitHub Actions environment to further restrict publishing to selected branches for additional security.
   - c.f. https://blog.pypi.org/posts/2023-04-20-introducing-trusted-publishers/
   - c.f. https://docs.pypi.org/trusted-publishers/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "github-actions"
+      - "dependencies"
+    reviewers:
+      - "asmeurer"
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -1,0 +1,105 @@
+name: publish distributions
+on:
+  push:
+    branches:
+    - main
+    tags:
+    - [0-9]+.[0-9]+
+    - [0-9]+.[0-9]+.[0-9]+
+  pull_request:
+    branches:
+    - main
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      publish:
+        type: choice
+        description: 'Publish to TestPyPI?'
+        options:
+        - false
+        - true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build:
+    name: Build Python distribution
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+
+    - name: Install python-build and twine
+      run: |
+        python -m pip install --upgrade pip setuptools
+        python -m pip install build twine
+        python -m pip list
+
+    - name: Build a wheel and a sdist
+      run: |
+        PYTHONWARNINGS=error,default::DeprecationWarning python -m build .
+
+    - name: Verify the distribution
+      run: twine check --strict dist/*
+
+    - name: List contents of sdist
+      run: python -m tarfile --list dist/array_api_compat-*.tar.gz
+
+    - name: List contents of wheel
+      run: python -m zipfile --list dist/array_api_compat-*.whl
+
+    - name: Upload distribution artifact
+      uses: actions/upload-artifact@v3
+      with:
+        name: dist-artifact
+        path: dist
+
+  publish:
+    name: Publish Python distribution to (Test)PyPI
+    if: github.event_name != 'pull_request' && github.repository == 'data-apis/array-api-compat'
+    needs: build
+    runs-on: ubuntu-latest
+    # Mandatory for publishing with a trusted publisher
+    # c.f. https://docs.pypi.org/trusted-publishers/using-a-publisher/
+    permissions:
+      id-token: write
+    # Restrict to the environment set for the trusted publisher
+    environment:
+      name: publish-package
+
+    steps:
+    - name: Download distribution artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: dist-artifact
+        path: dist
+
+    - name: List all files
+      run: ls -lh dist
+
+    - name: Publish distribution 📦 to Test PyPI
+      # Publish to TestPyPI on tag events of if manually triggered
+      # Compare to 'true' string as booleans get turned into strings in the console
+      if: >-
+        (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v'))
+        || (github.event_name == 'workflow_dispatch' && github.event.inputs.publish == 'true')
+      uses: pypa/gh-action-pypi-publish@v1.8.10
+      with:
+        repository-url: https://test.pypi.org/legacy/
+        print-hash: true
+
+    - name: Publish distribution 📦 to PyPI
+      if: github.event_name == 'release' && github.event.action == 'published'
+      uses: pypa/gh-action-pypi-publish@v1.8.10
+      with:
+        print-hash: true


matthewfeickert:
@asmeurer this is mostly ready to go, but before this gets finished I'll need you to add a trusted publisher to TestPyPI and PyPI with repo name `ata-apis/array-api-compat` and workflow name of `.github/workflows/publish-package.yml`.

https://docs.pypi.org/trusted-publishers/adding-a-publisher/

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,104 @@
+name: publish distributions
+on:
+  push:
+    # branches:
+    # - main
+    # tags:
+    # - v*
```
matthewfeickert, line 7: This was just commented out for me to verify things worked on my fork. I'll finish this soon and clean up and add context in PR body.

asmeurer:
> @asmeurer this is mostly ready to go, but before this gets finished I'll need you to add a trusted publisher to TestPyPI and PyPI with repo name ata-apis/array-api-compat and workflow name of .github/workflows/publish-package.yml.

I've done this. Should I add something for the environment name?

matthewfeickert:
> I've done this. Should I add something for the environment name?

Ah yeah, sorry I forgot to mention this. @asmeurer can you go into https://github.com/data-apis/array-api-compat/settings/environments and add an environment named `publish-package` with deployment branches that match to pasterns of `main`, `v*`, and whatever release branch names backports might be released against (e.g. `release/*`)? c.f. https://github.com/scikit-hep/pyhf/pull/2183 for lots of discussion on this. Let me know if you have any questions.

Screenshot from 2023-07-18 00-34-44

.github/dependabot.yml
matthewfeickert, line 1: This dependabot config is _only_ for updating GitHub Actions, and so the number of PRs generated should be substantially less than if Python packages were being considered.

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
...
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "github-actions"
+      - "dependencies"
+    reviewers:
+      - "asmeurer"
```
matthewfeickert, line 12: This assumes @asmeurer wants to be the person to get the ping.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+on:
+  push:
+    branches:
+    - main
+    tags:
+    - v*
+  pull_request:
+    branches:
+    - main
+    - release/v*
```
matthewfeickert, line 11: This is a assuming a release branch pattern for future backports for bug fixes to patch versions of `release/v*` which may not be the case. This can be altered as desired.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+  release:
+    types: [published]
```
matthewfeickert, line 13: This bit is for publishing on GitHub Release.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+  workflow_dispatch:
+    inputs:
+      publish:
+        type: choice
+        description: 'Publish to TestPyPI?'
+        options:
+        - false
+        - true
```
matthewfeickert, line 21: This would allow you to create a release to TestPyPI from workflow dispatch. If this isn't of interest this can get removed.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+        fetch-depth: 0
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+
+    - name: Install python-build and twine
+      run: |
+        python -m pip install --upgrade pip setuptools
```
matthewfeickert, line 44: If `array-api-compat` ever moves off `setuptools` this `setuptools` bit here could be removed.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+    - name: Publish distribution 📦 to Test PyPI
+      # Publish to TestPyPI on tag events of if manually triggered
+      # Compare to 'true' string as booleans get turned into strings in the console
+      if: >-
+        (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v'))
+        || (github.event_name == 'workflow_dispatch' && github.event.inputs.publish == 'true')
+      uses: pypa/gh-action-pypi-publish@v1.8.8
+      with:
+        repository-url: https://test.pypi.org/legacy/
+        print-hash: true
```
matthewfeickert, line 99: Publishing to TestPyPI (which is optional of course) would require following these instructions as there is no `array-api-compat` yet on TestPyPI: https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+
+    - name: Publish distribution 📦 to PyPI
+      if: github.event_name == 'release' && github.event.action == 'published'
```
matthewfeickert, line 102: This is assuming that you only want to publish to PyPI via creating a GitHub Release.

.github/workflows/publish-package.yml
asmeurer: I don't *think* we need to worry about backports here. I'm trying to think of a situation where we'd want to do that. 

We generally try to avoid making backwards incompatible changes to the standard itself, but even if there was one, I think we would want to support multiple versions in the same array-api-compat version using [`api_version`](https://github.com/data-apis/array-api-compat/blob/ea6a9d695d011ff84c0b1ebd1cb624875fe9d131/array_api_compat/common/_helpers.py#L56). Similarly, we'll probably want to support both NumPy 2.0 and NumPy 1.25 at the same time when that happens.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
+name: publish distributions
+on:
+  push:
+    branches:
+    - main
+    tags:
+    - v*
```
asmeurer, line 7: The tags on this repo don't have the v prefix.

matthewfeickert: I'll replace these with a regex later.

asmeurer: I pushed a fix for this.

.github/workflows/publish-package.yml
asmeurer: Is this the right link? I don't see testpypi mentioned on that page.

.github/workflows/publish-package.yml
asmeurer: This is worth doing at least the first time around. How do you do it? Is there some option when you create the release in the GitHub interface?

.github/workflows/publish-package.yml
matthewfeickert: Thanks @asmeurer. Sorry I'm currently traveling for work but will get back to this in the next 48 hours.

.github/workflows/publish-package.yml
matthewfeickert: > I don't _think_ we need to worry about backports here.

Sounds good. I'll drop this then if you're not planning on making releases from any branch other than `main`.

.github/workflows/publish-package.yml
matthewfeickert: You do this through workflow dispatch by visiting the GitHub Actions tab for the repo and then selecting this workflow (https://github.com/data-apis/array-api-compat/actions/workflows/publish-package.yml) and then when you click the "Run workflow" button on the right hand side there will be the option to select `false` or `true` from a dropdown.

This is how it looks like for us on `pyhf`'s GitHub repo:

pyhf-view

and if it helps, starting around 52 seconds into this video of me making the `pyhf` `v0.7.1` release I show this button example: https://youtu.be/ZV20tr3EpTw?t=52

.github/workflows/publish-package.yml
matthewfeickert: That is the right link. Both PyPI and TestPyPI have the setup for trusted publishing in the same area. So just treat all instances of PyPI as TestPyPI for creating an project with a Trusted Publishers workflow.

.github/workflows/publish-package.yml
matthewfeickert: This has now been removed.

matthewfeickert:
Sorry for the 2 week delay &mdash; I was traveling for work and then finishing faculty applications &mdash; but this should be ready for review again.

asmeurer:
Thanks for the update. It might be another 2 weeks before I myself am able to look at this.

.github/workflows/publish-package.yml
matthewfeickert: bumped this from `pypa/gh-action-pypi-publish@v1.8.8` to `pypa/gh-action-pypi-publish@v1.8.10`.

asmeurer:
We need to do a release so I'm going to merge this and see if it works.

matthewfeickert:
> We need to do a release so I'm going to merge this and see if it works.

@asmeurer let me know how this goes. Tag based deployments under environments have started breaking recently given https://github.com/orgs/community/discussions/62991, so for the most recent release I've made (https://github.com/scikit-hep/pyhf/issues/2319) I had to temorarily set the deployment branches in the "publish-package" environment to "all branches", which invalidates https://github.com/data-apis/array-api-compat/pull/51#issuecomment-1639518801 a bit.

.github/workflows/publish-package.yml
```diff
@@ -0,0 +1,105 @@
...
+        path: dist
+
+    - name: List all files
+      run: ls -lh dist
+
+    - name: Publish distribution 📦 to Test PyPI
+      # Publish to TestPyPI on tag events of if manually triggered
+      # Compare to 'true' string as booleans get turned into strings in the console
+      if: >-
+        (github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v'))
```
matthewfeickert, line 94: @asmeurer Also catching now that as you use tags that match

```
    - [0-9]+.[0-9]+
    - [0-9]+.[0-9]+.[0-9]+
```

but not `v*` that these will never publish to TestPyPI on tag. I should fix this in a follow up PR.

Issue #54, opened 2023-08-18
BUG: common: `linalg.cholesky` returns the conjugate of the expected upper decomposition


lucascolley:
The same as https://github.com/numpy/numpy/issues/24451, but for `array_api_compat.cupy`.

If I am correct about this being a bug, then I suspect that this is the line causing it:
https://github.com/data-apis/array-api-compat/blob/main/array_api_compat/common/_linalg.py#L58C47-L58C47

PR #61, opened 2023-09-22
BUG: common: fix cholesky upper decomp for complex dtypes

lucascolley:
Closes gh-54.

cc @asmeurer

### DIFF
--- a/array_api_compat/common/_linalg.py
+++ b/array_api_compat/common/_linalg.py
@@ -11,7 +11,7 @@ if np.__version__[0] == "2":
 else:
     from numpy.core.numeric import normalize_axis_tuple
 
-from ._aliases import matmul, matrix_transpose, tensordot, vecdot
+from ._aliases import matmul, matrix_transpose, tensordot, vecdot, isdtype
 from .._internal import get_xp
 
 # These are in the main NumPy namespace but not in numpy.linalg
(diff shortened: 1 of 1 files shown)


rgommers (approved):
Thanks @lucascolley! I verified the change on your NumPy PR, and this matches that one. In it goes.

rgommers:
Actually, let me re-run CI since a PyTorch job was red and I couldn't see the CI log anymore.

lucascolley:
thanks Ralf. I wonder whether any changes are needed to `array-api-tests` since this wasn't caught initially?

rgommers:
> I wonder whether any changes are needed to `array-api-tests` since this wasn't caught initially?

I checked quickly, and yes indeed - `upper` is tested, but only with real-valued dtypes (via `floating_dtypes` from hypothesis). I guess the fix is to use `complex_dtypes` from https://hypothesis.readthedocs.io/en/latest/_modules/hypothesis/extra/array_api.html. Are you interested to follow up on that?

rgommers:
Several failures due to `fft` testing mainly, I assume because new tests were added in the test suite and not yet here.

All the `linalg.cholesky` tests passed, so I'll go ahead and merge this.

lucascolley:
> Are you interested to follow up on that?

I had a look but I'm not sure how to fix. Will open an issue instead.

Issue #64, opened 2023-10-23
AttributeError: module 'array_api_compat.torch' has no attribute 'newaxis'


ogrisel:
According to:

- https://data-apis.org/array-api/latest/API_specification/generated/array_api.constants.newaxis.html

the namespace modules should expose an alias to `None` named `newaxis` to make indexing code more explicit.

However it seems to be missing in `array-api-compat`, at least for PyTorch:

```python
>>> import array_api_compat.torch as xp
>>> xp.newaxis is None
Traceback (most recent call last):
  Cell In[6], line 1
    xp.newaxis
AttributeError: module 'array_api_compat.torch' has no attribute 'newaxis'
```

Note, the same snippet works with numpy:

```python
>>> import array_api_compat.numpy as xp
>>> xp.newaxis is None
True
```

asmeurer:
I guess newaxis isn't in the main torch namespace. I would have expected this to be caught by the test suite, but I guess not.

Issue #65, opened 2023-11-09
Test failure for `test_multiply` in pytorch-latest CI job


rgommers:
Observed in gh-59 and gh-61 with the same test case:
`array_api_tests/test_operators_and_elementwise_functions.py::test_multiply[__imul__(x1, x2)] `
just with different values. The reproducers:

```
FAILED array_api_tests/test_operators_and_elementwise_functions.py::test_multiply[__imul__(x1, x2)] - AssertionError: x1=(2+67138304j), but should be roughly (x1 * x2)=67138308j [__imul__()]
  x1=(9638+9638j), x2=(3483+3483j)
assert False
 +  where False = isclose(2.0, 0.0, 3.4028234663852886e+38)
 +    where 2.0 = (2+67138304j).real
 +    and   0.0 = 67138308j.real
Falsifying example: test_multiply(
    ctx=BinaryParamContext(<__imul__(x1, x2)>),
    data=data(...),
)
Draw 1 (x1): tensor(9638.+9638.j)
Draw 2 (x2): tensor(3483.+3483.j)

You can reproduce this example by temporarily adding @reproduce_failure('6.86.2', b'AXicY2eAA9Vl2JgMvLORmQA8GwLu') as a decorator on your test case
```

and

```
FAILED array_api_tests/test_operators_and_elementwise_functions.py::test_multiply[__imul__(x1, x2)] - AssertionError: x1=(-5+92034536j), but should be roughly (x1 * x2)=(-3+92034537j) [__imul__()]
  x1=(15339089+15339090j), x2=(3+3j)
assert False
 +  where False = isclose(-5.0, -3.0, 3.4028234663852886e+38)
 +    where -5.0 = (-5+92034536j).real
 +    and   -3.0 = (-3+92034537j).real
Falsifying example: test_multiply(
    ctx=BinaryParamContext(<__imul__(x1, x2)>),
    data=data(...),
)
Draw 1 (x1): tensor(15339089.+15339090.j)
Draw 2 (x2): tensor(3.+3.j)

You can reproduce this example by temporarily adding @reproduce_failure('6.86.1', b'AXicY2eAgVd8gQhmEAMCMCMzAVPvAqE=') as a decorator on your test case
```

A comment on gh-59 has a full traceback.

asmeurer:
Probably the same sort of issue discussed at https://github.com/data-apis/array-api-tests/issues/190. The test suite isn't checking complex number closeness correctly.

[...]

PR #66, opened 2023-11-09
Add `newaxis` to torch aliases

rgommers:
Closes gh-64

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -144,6 +144,7 @@ def can_cast(from_: Union[Dtype, array], to: Dtype, /) -> bool:
 
 # Basic renames
 bitwise_invert = torch.bitwise_not
+newaxis = None
 
 # Two-arg elementwise functions
 # These require a wrapper to do the correct type promotion on 0-D tensors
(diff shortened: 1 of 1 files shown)


rgommers:
Will add a corresponding test case to `array-api_tests` as well.

asmeurer:
Looks good. Merging might be blocked on #65 though.

array_api_compat/torch/_aliases.py
```diff
@@ -694,7 +695,7 @@ def take(x: array, indices: array, /, *, axis: Optional[int] = None, **kwargs) -
            'atan2', 'bitwise_and', 'bitwise_left_shift', 'bitwise_or',
            'bitwise_right_shift', 'bitwise_xor', 'divide', 'equal',
            'floor_divide', 'greater', 'greater_equal', 'less', 'less_equal',
-           'logaddexp', 'multiply', 'not_equal', 'pow', 'remainder',
+           'logaddexp', 'multiply', 'newaxis', 'not_equal', 'pow', 'remainder',
```
asmeurer, line 698: I've tried to keep these in the same order they are in the file (although I see now that that isn't really completely true).

rgommers:
This was green before, and gh-65 is unrelated and should not be blocking. No idea why the `fft` tests are failing now, probably another case of "should use derandomize". Either way, let's merge this?

asmeurer:
We'll have to fix the CI before we release at least.

Issue #68, opened 2023-11-17
Improve error message for invalid input


vnmabus:
Right now, the error message passing an invalid object to `array_namespace` is 
```
The input is not a supported array type
```
It would be more useful for debugging to say which class failed, I think. E.g.:
```
"generator" is not a supported array type
```
```
"map" is not a supported array type
```
```
"float" is not a supported array type
```

PR #72, opened 2023-12-11
MAINT: improve error message for `array_namespace`

lucascolley:
Closes gh-68

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -99,7 +99,7 @@ def array_namespace(*xs, api_version=None, _use_compat=True):
                 namespaces.add(torch)
         else:
             # TODO: Support Python scalars?
-            raise TypeError(f"{type(x)!r} is not a supported array type")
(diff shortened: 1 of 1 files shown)


array_api_compat/common/_helpers.py
```diff
@@ -99,7 +99,7 @@ def your_function(x, y):
                 namespaces.add(torch)
         else:
             # TODO: Support Python scalars?
-            raise TypeError("The input is not a supported array type")
+            raise TypeError(f"{type(x)!r} is not a supported array type")
```
asmeurer, line 102: `repr(type(x))` is like `<class 'int'>`. You want something like `type(x).__name__`.

lucascolley:
```suggestion
            raise TypeError(f"{type(x).__name__} is not a supported array type")
```

rgommers (approved):
LGTM, and this 1 line change adopts the suggestion from @asmeurer's review comment - so in it goes. Thanks @lucascolley

PR #76, opened 2023-12-19
Add dask to array-api-compat

lithomas1:


lithomas1:
Dask tests should be green, but very slow now.

array_api_compat/_dask_ci_shim.py
```diff
@@ -0,0 +1,13 @@
+"""
+A little CI shim for the dask backend that
+disables the dask scheduler
+"""
+import dask
+dask.config.set(scheduler='synchronous')
```
lithomas1, line 6: I don't know if this is needed anymore (it didn't seem to help with the test that OOMed - which is now disabled).

Leaving this in just in case I haven't found the correct options.

array_api_compat/common/_aliases.py
```diff
@@ -303,6 +303,8 @@ def _asarray(
         import numpy as xp
     elif namespace == 'cupy':
         import cupy as xp
+    elif namespace == 'dask':
```
lithomas1, line 306: Technically, the array module of dask is called ``dask.array``.

I guess we could match that here, but it feels kind of weird to me.

array_api_compat/common/_helpers.py
```diff
@@ -219,6 +236,13 @@ def to_device(x: "Array", device: "Device", /, *, stream: "Optional[Union[int, A
         return _cupy_to_device(x, device, stream=stream)
     elif _is_torch_array(x):
         return _torch_to_device(x, device, stream=stream)
+    elif _is_dask_array(x):
+        if stream is not None:
+            raise ValueError("The stream argument to to_device() is not supported")
+        # TODO: What if our array is on the GPU already?
```
lithomas1, line 242: I think dask supports cupy on the GPU.

Is this something we also need to take into consideration?

lithomas1:
@rgommers 
This should be ready for a first pass now.

array_api_compat/common/_helpers.py
rgommers: Yes indeed, it does support CuPy. I"m not sure that a `stream` input can be supported though. If a Dask array spans multiple machines, I think there'd be multiple streams with numbers that are unrelated to each other. Which therefore can't be supported at the Dask level in this API.

That's probably fine - you'd only move arrays in a single process to another device like this I think, so maybe the whole `to_device` method doesn't quite work for Dask? @jakirkham any thoughts on this?

array_api_compat/common/_aliases.py
rgommers: I think (but am not 100% sure) that using `namespace == 'dask.array'` here is preferred here, since it's meant to identify the actual namespace rather than the package.

rgommers: I suggest waiting for @asmeurer to weigh in here.

rgommers:
Nice @lithomas1! Question: did you already try this with scikit-learn or SciPy functionality that uses `array_api_compat` under the hood, and if so did that work?

> Array API Tests (Dask) ... Failing after 130m 

Let's try and fix that first I'd say, this isn't really workable. It uses `max_examples=200` says the log - how about starting with `1`? The first goal here is checking what exists yes/no, for which you need only a single example. That will make it much easier to iterate on this. I assume that'll fix this practical issue - maybe @honno or @asmeurer have better ideas.

lithomas1:
Thanks for the review.

> Nice @lithomas1! Question: did you already try this with scikit-learn or SciPy functionality that uses `array_api_compat` under the hood, and if so did that work?

I'm currently testing with scikit-learn, and the biggest issue I'm facing is the array having NaN shapes (since the dask computation hasn't been compute()'ed yet). From a quick glance, it looks like scikit-learn uses the shape in allocating the result array, so there doesn't seem to be easy way to avoid the shape call.

Currently, I'm working around that by adding a lot of ``np.asarray`` calls to force computation - but that probably lowers performance for the other eager backends, right?

Any ideas on how to fix this?

> > Array API Tests (Dask) ... Failing after 130m
> 
> Let's try and fix that first I'd say, this isn't really workable. It uses `max_examples=200` says the log - how about starting with `1`? The first goal here is checking what exists yes/no, for which you need only a single example. That will make it much easier to iterate on this. I assume that'll fix this practical issue - maybe @honno or @asmeurer have better ideas.

Hm, I think I got this to green last time, even with the full amount of max-examples.

I'm going to try lowering it to 5. 
(I remember that when testing this last time, there were tests that failed when you increased the number of max-examples)

lithomas1:
Update on the testing with scikit-learn/scipy:

I did a really quick test with scikit-learns ``train_test_split`` function, and it looks like that worked.

rgommers:
> I did a really quick test with scikit-learns `train_test_split` function, and it looks like that worked.

Nice! That's probably the most heavily used function in all of scikit-learn, so a good start:)

> and the biggest issue I'm facing is the array having NaN shapes (since the dask computation hasn't been compute()'ed yet). From a quick glance, it looks like scikit-learn uses the shape in allocating the result array, so there doesn't seem to be easy way to avoid the shape call.
> 
> Currently, I'm working around that by adding a lot of `np.asarray` calls to force computation - but that probably lowers performance for the other eager backends, right?
> 
> Any ideas on how to fix this?

Yes, not by much but there is overhead for each `asarray` call. In general it's tricky. `nan` shapes should come from data-dependent operations mostly, so I guess you're running into something like this?

```python
>>> x = da.ones((3))
>>> y = da.unique(da.zeros((3)))
>>> x
dask.array<ones_like, shape=(3,), dtype=float64, chunksize=(3,), chunktype=numpy.ndarray>
>>> y
dask.array<getitem, shape=(nan,), dtype=float64, chunksize=(nan,), chunktype=numpy.ndarray>
>>> da.shape(x + y)
...
ValueError: operands could not be broadcast together with shapes (3,) (nan,)
```

Allocating new arrays with a shape that depends on the result shape of some other operation with a `nan` in the shape attribute may be tricky. In some cases it does work, e.g.:
```python
>>> z = da.empty_like(y)
>>> z
dask.array<empty_like, shape=(nan,), dtype=float64, chunksize=(nan,), chunktype=numpy.ndarray>
```
and that can stay lazy just fine. But if it's not the same shape and you need a new array with an extra dimension, things become tricky I guess, since the `nan` is actually a `nan` and not some custom object that can be carried forward when constructing a new array.

Can you point to the code in question?

lithomas1:
This one is the one that I've hit so far.

https://github.com/scikit-learn/scikit-learn/blob/e2b3785b6a1f96989c3992bffa2a05ef5c048a7e/sklearn/discriminant_analysis.py#L116

Not too familiar with the sklearn code, but would it be fair to assume that a good chunk of places where we have ``np.zeros`` would be converted to ``xp.zeros`` (as the code is made array API compatible)?

Then we would have a lot more 
https://github.com/search?q=repo%3Ascikit-learn%2Fscikit-learn+np.zeros+language%3APython+path%3A%2F%5Esklearn%5C%2F%2F&type=code

rgommers:
> Not too familiar with the sklearn code, but would it be fair to assume that a good chunk of places where we have `np.zeros` would be converted to `xp.zeros` (as the code is made array API compatible)?

Summary of what we just discussed:
- Dask is able to do shape computations in many cases, so we don't expect too many cases where this `nan` issue will be problematic.
- For those cases, it's probably fine to write a small utility function for use in scikit-learn that checks if an array is a dask array and if so returns `.compute()` and otherwise is a no-op.

betatim:
Joining in for the chat about scikit-learn. In the particular case you [linked to](https://github.com/scikit-learn/scikit-learn/blob/e2b3785b6a1f96989c3992bffa2a05ef5c048a7e/sklearn/discriminant_analysis.py#L116) the code is computing something from the input data (the number of classes) and then creating a new array with a shape based on the number of classes found in the input data.

I think this means there isn't much we can do, other than getting an actual value for the shape. Ideally the underlying lib (dask in this case) can compute that without doing all of the outstanding work. It sounds like that is what dask is (sometimes) doing?

I'm not sure what I think about adding code to scikit-learn to deal with "if dask then call `.compute()` else no-op". The whole point of the Array API is that as a array consumer you don't have to know the producer of the array. It is annoying to have that kind of code, but that isn't the reason I dislike it. The reason I dislike it is that it means we can't write "scikit-learn supports all Array API compliant arrays as input". Instead we need to explicitly list the set of Array API compliant inputs that scikit-learn handles.

I'm not sure what the right solution is. It is probably bigger than array-api-compat though. In the discussions about lazy vs not-lazy array producers my impression was that we kinda didn't really conclude anything. People like me would like accessing things like `.shape` implicitly to trigger a computation (if the underlying array producer is lazy), the lazy library implementors prefer explicit `.compute()` calls or the concept of a `LazyInteger` (I think).

So I don't know how/where/if we should restart this discussion :-/

rgommers:
> People like me would like accessing things like `.shape` implicitly to trigger a computation (if the underlying array producer is lazy), the lazy library implementors prefer explicit `.compute()` calls or the concept of a `LazyInteger` (I think).

To me this feels more like missing functionality in Dask. Accessing `.shape[0]` to use as input to constructing a new array can stay lazy just fine - if only dask wouldn't use plain `nan` but a smarter object to represent "not yet computed. It could have a repr of `nan` while being an object that carries a reference to its parent array's shape. This already works if you use the whole shape (see `z = da.empty_like(y)` above), it's just not implemented for a partial use of that shape. And having that functionality should be useful for many such cases.

@jakirkham do you think this is implementable in Dask with reasonable effort?

array_api_compat/common/_aliases.py
asmeurer: This keyword is only used internally so it doesn't really matter, but I like `dask.array`.

array_api_compat/common/_aliases.py
```diff
@@ -322,10 +324,13 @@ def _asarray(
     if copy in COPY_FALSE:
         # copy=False is not yet implemented in xp.asarray
         raise NotImplementedError("copy=False is not yet implemented")
-    if isinstance(obj, xp.ndarray):
+    # TODO: This feels wrong (__array__ is not in the standard)
+    # Dask doesn't support DLPack, though, so, this'll do
```
asmeurer, line 328: This function is only used for numpy and cupy (and now dask). The internals of the wrappers can use specific functionality that isn't in the standard because they can assume that they are operating on the specific array library. So this isn't a problem.

array_api_compat/common/_aliases.py
```diff
@@ -322,10 +324,13 @@ def _asarray(
...
+    if (hasattr(xp, "ndarray") and isinstance(obj, xp.ndarray)) or hasattr(obj, "__array__"):
         if dtype is not None and obj.dtype != dtype:
             copy = True
-        if copy in COPY_TRUE:
+        # Dask arrays are immutable, so copy doesn't do anything
```
asmeurer, line 332: I think this is probably fine but maybe this sort of thing should be clarified in the standard.

array_api_compat/common/_helpers.py
asmeurer: Even if dask can't really support device transfers or anything I would not call the default device "cpu" as that's misleading. We could just create a proxy `DaskDevice` object that serves as the device for dask arrays.

asmeurer:
This is looking good. It looks like the basis for most of this is that the dask functions are basically the same as numpy, so the existing wrappers that are used for numpy and cupy can just be reused for dask (i.e., all of [these](https://github.com/data-apis/array-api-compat/pull/76/files#diff-fd9af9c46a5a0065749c873fb1aae2a3f3500b1ce738491b8f66354758e8db15R80-R102). I'll assume that you've actually verified that this is the case for these functions either manually or using the tests. 

Note that if a some dask function does end up deviating significantly from numpy, don't feel like you need to always use the helpers in `common`. Those are there just to avoid code duplicate between numpy and cupy, but for instance, you'll notice that the `torch` wrappers [barely use common at all](https://github.com/data-apis/array-api-compat/blob/main/array_api_compat/torch/_aliases.py).

array_api_compat/common/_aliases.py
lithomas1: Yeah, I think what I was mostly worried about was whether I need to return a new dask array object (I guess this would kinda be like a view on the old object) or not.

lithomas1:
> This is looking good. It looks like the basis for most of this is that the dask functions are basically the same as numpy, so the existing wrappers that are used for numpy and cupy can just be reused for dask (i.e., all of [these](https://github.com/data-apis/array-api-compat/pull/76/files#diff-fd9af9c46a5a0065749c873fb1aae2a3f3500b1ce738491b8f66354758e8db15R80-R102). I'll assume that you've actually verified that this is the case for these functions either manually or using the tests.

I just finished wrapping the linalg functions, but I haven't done the wrapping on the FFT functions at all, though.
Should I wait for #78 to be merged first, then?

lithomas1:
@asmeurer @rgommers 

This should be ready for a final pass now.

dask-xfails.txt
```diff
@@ -0,0 +1,139 @@
...
+# a = da.array([1]).reshape((1,1))
+# key = (0, slice(None, None, -1))
+# a[key] = da.array([1])
+
+# Failing hypothesis test case
+#x=dask.array<zeros_like, shape=(0, 2), dtype=bool, chunksize=(0, 2), chunktype=numpy.ndarray>
+#| Draw 1 (key): (slice(None, None, None), slice(None, None, None))
+#| Draw 2 (value): dask.array<zeros_like, shape=(0, 2), dtype=bool, chunksize=(0, 2), chunktype=numpy.ndarray>
+
+# TODO: this also skips test_setitem_masking unnecessarily
```
asmeurer, line 12: Let's change the xfails file matching to require an exact match. No reason not to do that.

lithomas1: Should I make a PR into array-api-tests?

asmeurer: Yes go ahead. It should be a trivial change. Just change some `in`s to `==` in conftest.py, and make sure some existing xfail files don't need to be updated.

asmeurer:
The code itself here looks fine. My main comment is the same as before, that we should be sure that the numpy/cupy generic wrappers are all actually correct for dask too. But the test suite will also be a good check on this. Have you ran the tests with more than the 5 max examples that are used on the CI? 

I guess the best thing to do would be to test this out with some consumer library like scipy or scikit-learn. If it's easier to do that after making a release we can do that. 

Should we also see if any dask devs want to take a look here?

array_api_compat/common/_helpers.py
asmeurer: @lithomas1 what's the status of this comment? Does the `device()` helper above need to be updated for dask?

array_api_compat/common/_linalg.py
```diff
@@ -77,7 +77,8 @@ def matrix_rank(x: ndarray,
     # dimensional arrays.
     if x.ndim < 2:
         raise xp.linalg.LinAlgError("1-dimensional array given. Array must be at least two-dimensional")
-    S = xp.linalg.svd(x, compute_uv=False, **kwargs)
+    S = xp.linalg.svdvals(x, **kwargs)
+    #S = xp.linalg.svd(x, compute_uv=False, **kwargs)
```
asmeurer, line 81: This change looks like it is causing the numpy array api tests to fail.

array_api_compat/dask/array/_aliases.py
```diff
@@ -0,0 +1,123 @@
...
+    arctanh as atanh,
+    left_shift as bitwise_left_shift,
+    right_shift as bitwise_right_shift,
+    invert as bitwise_invert,
+    power as pow,
+    # Other
+    concatenate as concat,
+)
+
+del da
```
asmeurer, line 123: Let's also manage `__all__` in these files the same as in the other submodules.

array_api_compat/dask/array/linalg.py
```diff
@@ -0,0 +1,42 @@
...
+    from ...common._typing import ndarray, Device, Dtype
+
+#cross = get_xp(da)(_linalg.cross)
+#outer = get_xp(da)(_linalg.outer)
+EighResult = _linalg.EighResult
+QRResult = _linalg.QRResult
+SlogdetResult = _linalg.SlogdetResult
+SVDResult = _linalg.SVDResult
+qr = get_xp(da)(_linalg.qr)
+#svd = get_xp(da)(_linalg.svd)
```
asmeurer, line 23: Can you get rid of all the little comment out bits of code like this. Or if one of them actually is useful to keep, add an explanatory comment above it.

asmeurer:
> Should we also see if any dask devs want to take a look here?

We should also see if the dask devs are interested in having upstream issues for the existing holes here. 

@lithomas1 if any upstream issues do exist, or are opened, please cross-reference them in the xfails file (and the code if relevant).

array_api_compat/common/_linalg.py
asmeurer: The current numpy dev CI failures can be ignored (they should go away once https://github.com/numpy/numpy/pull/25668 is merged), but the other ones are important.

lithomas1: Thanks for the catch. I think this should be equal to svdvals (if svdvals exists), so I just used a hasattr check. Lemme know if that's too hacky

array_api_compat/dask/array/_aliases.py
lithomas1: Updated for everything except linalg.

I'm not sure if we can do linalg since there's no __all__ for dask.array.linalg.

dask-xfails.txt
lithomas1: So, it turns out that after fixing some other tests, this test(``test_setitem_masking``) started failing again (the test was passing because of a bug I had in my asarray version), and the comment isn't relevant anymore.

lithomas1:
> > Should we also see if any dask devs want to take a look here?
> 
> We should also see if the dask devs are interested in having upstream issues for the existing holes here.

Definitely. I'll try to get followups opened up on the dask side after this is merged (so I don't have to go back and forth if new tests fail/pass).

array_api_compat/common/_linalg.py
asmeurer: You should use `get_xp(xp).linalg.svdvals` so that it uses the svdvals alias. This pattern is already used in a few places in this file. 

As for not wrapping svdvals when it is present, we could do that too (I did this for `vector_norm` https://github.com/data-apis/array-api-compat/blob/916a84bafd2ae97c09d012616a13ee03354a1480/array_api_compat/numpy/linalg.py#L30-L33 but I forgot about svdvals).

array_api_compat/dask/array/_aliases.py
asmeurer: See how it's done for cupy.linalg, which also doesn't define `__all__`

array_api_compat/dask/array/linalg.py
lithomas1: Sorry, had some code left over from debugging.

Everything should be cleaned up now. 🤞

array_api_compat/common/_linalg.py
lithomas1: Thanks, updated.

array_api_compat/dask/array/_aliases.py
lithomas1: This should be updated as well.

asmeurer:
Alright, I'm going to merge this. I'll make a release with this and if any further issues come up we can fix them up later. I don't think anyone will test this unless it is released.

asmeurer:
It also looks like the `device` helper doesn't work for dask. I added a test at #84 but XFAILed it. Can you make a PR fixing this?

lithomas1:
Will do. Feel free to put in the changes from #84 first (It'll probably be easier for me to rebase on top of that).

asmeurer:
@lithomas1 I just merged #82 which updates some of your dask code. Can you make a PR fixing these dask issues? There are also several test failures on CI related to dask. #84 shouldn't be an issue. Just add support for `device` but don't worry about testing it.

lithomas1:
Will do, was slightly busy yesterday, but should have both PRs up tomorrow.

PR #80, opened 2024-01-18
Fixing CI test failures

asmeurer:
- Added a torch wrapper for `trace` (`torch.trace` doesn't support the `offset` argument or stacking)
- Wrap numpy, cupy, and torch `nonzero` to raise an error for zero-dimensional input arrays.


asmeurer:
I don't believe it, but the CI is actually green.

rgommers:
Nice work!

PR #81, opened 2024-01-19
1.4.1 release

asmeurer:
Need to check the cupy tests first.


asmeurer:
CuPy tests are passing. NumPy dev is just failing because the dev builds for numpy are lagging. I just verified manually that they work with numpy main.

PR #84, opened 2024-02-06
Add basic JAX support

asmeurer:
Unlike other modules, JAX array API support is fully in JAX itself in the `jax.experimental.array_api` submodule, so the only thing that is done here is to add JAX support to the helper functions. This also means that we do not run array-api-tests on JAX.

This also makes the various `is_numpy_array`, `is_cupy_array`, etc. functions public, as I noticed someone was using them on GitHub and they seem like they could be useful.


array_api_compat/common/_helpers.py
asmeurer, line 1: @jakevdp, I could mostly use your review for the changes in this file.

array_api_compat/common/_helpers.py
```diff
@@ -81,37 +92,44 @@ def your_function(x, y):
...
                 from ..dask import array as dask_namespace
                 namespaces.add(dask_namespace)
             else:
                 raise TypeError("_use_compat cannot be False if input array is a dask array!")
+        elif is_jax_array(x):
+            _check_api_version(api_version)
+            # jax.numpy is already an array namespace, but requires this
+            # side-effecting import for __array_namespace__ and some other
+            # things to be defined.
+            import jax.experimental.array_api as jnp
```
asmeurer, line 131: Will this import go away at some point? Should we guard against that?

array_api_compat/common/_helpers.py
```diff
@@ -156,8 +174,17 @@ def device(x: "Array", /) -> "Device":
...
         a ``device`` object (see the "Device Support" section of the array API specification).
     """
-    if _is_numpy_array(x):
+    if is_numpy_array(x):
         return "cpu"
+    if is_jax_array(x):
+        # JAX has .device() as a method, but it is being deprecated so that it
+        # can become a property, in accordance with the standard. In order for
+        # this function to not break when JAX makes the flip, we check for
+        # both here.
```
asmeurer, line 183: Does this logic seem OK?

array_api_compat/common/_helpers.py
```diff
@@ -204,6 +231,12 @@ def _torch_to_device(x, device, /, stream=None):
         raise NotImplementedError
     return x.to(device)
 
+def _jax_to_device(x, device, /, stream=None):
+    import jax
+    if stream is not None:
+        raise NotImplementedError
+    return jax.device_put(x, device)
```
asmeurer, line 238: Is this helper function for `to_device` correct?

array_api_compat/common/_helpers.py
jakevdp: It's not true that `jax.numpy` is already an array namespace. `jax.experimental.array_api` is an array namespace, and we hope to make `jax.numpy` an array namespace in the future.

array_api_compat/common/_helpers.py
```diff
@@ -49,14 +50,24 @@ def _is_dask_array(x):
...
     return isinstance(x, dask.array.Array)
 
+def is_jax_array(x):
+    # Avoid importing jax if it isn't already
+    if 'jax' not in sys.modules:
+        return False
+
+    import jax.numpy
+
+    return isinstance(x, jax.numpy.ndarray)
```
jakevdp, line 60: `isinstance(x, jax.Array)` would be more concise.

asmeurer:
Oh we should also probably add a basic test for the device helpers at https://github.com/data-apis/array-api-compat/blob/main/tests/test_common.py

array_api_compat/common/_helpers.py
jakevdp: This looks OK, but it will not work with the `to_device` function as currently defined in `jax.experimental.array_api`. That expects to be passed the bound method object – it's a hack, and one of the reasons that this is still considered experimental.

array_api_compat/common/_helpers.py
asmeurer: OK, I fixed the comment. The question still remains though? Should I add a guard here like

```py
try:
    import jax.experimental.array_api as jnp
except ImportError:
    import jax.numpy as jnp
```

for a future JAX version when `jax.experimental.array_api` gets removed? Or will it never be removed?

array_api_compat/common/_helpers.py
jakevdp: The other problem here is that, in general, JAX arrays can live on multiple devices (in which case `arr.device()` will error). It's not clear from the Array API design how multi-device arrays are meant to be supported.

array_api_compat/common/_helpers.py
jakevdp: Yes, I think that's probably a reasonable way to future-proof this.

array_api_compat/common/_helpers.py
asmeurer: It looks like both forms work

```py
>>> import jax.experimental.array_api as xp
>>> x = xp.asarray([1, 2, 3])
>>> x.to_device(x.device)
Array([1, 2, 3], dtype=int32)
>>> x.to_device(x.device())
Array([1, 2, 3], dtype=int32)
```

asmeurer:
Apparently we've been supporting `"cpu"` as a special host device here (https://github.com/data-apis/array-api-compat/pull/40). This is still being discussed for the standard https://github.com/data-apis/array-api/issues/626. 

I'm actually unsure if it's a good idea for us to be supporting that here given that we haven't really agreed about it in the standard. But if we did want to support it for JAX, how would we? I don't see how to actually access `jax.CpuDevice`.

array_api_compat/common/_helpers.py
jakevdp: Yes, `device_put` would be the right method here.

jakevdp:
`jax.devices('cpu')[0]` would return the first CPU device.

array_api_compat/common/_helpers.py
```diff
@@ -55,9 +55,9 @@ def is_jax_array(x):
     if 'jax' not in sys.modules:
         return False
 
-    import jax.numpy
+    import jax
 
-    return isinstance(x, jax.numpy.ndarray)
+    return isinstance(x, jax.Array)
```
jakevdp, line 60: Should this be more guarded? e.g. what if someone has a module named `jax.py` in their path, it could lead to an `AttributeError` here and make `array_api_compat` unusable.

jakevdp (commented):
I think this looks good!

array_api_compat/common/_helpers.py
asmeurer: I don't know. That sort of thing tends to just break everything anyway. I've never really felt that libraries should protect against that sort of thing. 

Anyway, the whole point of this function is to be guarded. It won't import `jax` unless it's already been imported, because jax is a slow import and unnecessary if a user is using another array library.

array_api_compat/common/_helpers.py
asmeurer: The downside of that particular line of code is it will also pass through jax.numpy for older JAX versions that don't have jax.experimental.array_api. I like to avoid explicit version checks in this library if I can, but maybe that's the best thing to do here. 

Or maybe we can just change the logic to this once JAX starts to remove (deprecates?) the experimental import.

array_api_compat/common/_helpers.py
jakevdp: More precisely, it won't import `jax` unless *something* named `jax` has been imported, and then `array_api_compat` will irrecoverably error if that `jax` isn't the one we're expecting.

Same for every other package name referenced in this module.

My feeling is: it costs virtually nothing to wrap this all in an appropriate `try ... except`, so why not do so?

asmeurer:
Removed the "cpu" device logic. It looks like that isn't going to be part of the standard. I opened https://github.com/data-apis/array-api-compat/issues/86 about removing it for cupy as well.

array_api_compat/common/_helpers.py
asmeurer: I'm still not so sure this is a good idea. Usually if you have that sort of thing it will be an error for a lot of things, not just array_api_compat. My worry here is that guarding isn't as straightforward as it might seem. Wrapping everything in try/except could mean we end up silencing legitimate errors.

jakevdp: OK, sounds good.

PR #88, opened 2024-02-09
Update the api_version check to allow 2022.12 and warn on 2021.12

asmeurer:
Not sure if a warning is the best option here. The returned namespace is 2022.12 compliant, and that is also 2021.12 compliant (except for a few minor things). I don't think it's worth trying to support both, unless someone has a real use-case.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -16,6 +16,7 @@ if TYPE_CHECKING:
 import sys
 import math
 import inspect
+import warnings
 
 def is_numpy_array(x):
     # Avoid importing NumPy if it isn't already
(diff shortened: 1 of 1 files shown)


rgommers:
I'd leave out the warning and just return the superset of 2022.12

array_api_compat/common/_helpers.py
```diff
@@ -77,8 +78,10 @@ def is_array_api_obj(x):
         or hasattr(x, '__array_namespace__')
 
 def _check_api_version(api_version):
-    if api_version is not None and api_version != '2021.12':
-        raise ValueError("Only the 2021.12 version of the array API specification is currently supported")
+    if api_version == '2021.12':
+        warnings.warn("The 2021.12 version of the array API specification was requested but the returned namespace is actually version 2022.12")
+    if api_version is not None and api_version != '2022.12':
+        raise ValueError("Only the 2022.12 version of the array API specification is currently supported")
```
betatim, line 84: The text of the error isn't right, no? Because 2022.12 and 2021.12 are supported

---

Can we somehow re-do the `if` statements? I think something like the below would be easier to read and implement the behaviour Ralf suggested (no warning, just return subset)
```python
if api_version is not in ('2021.12', '2022.12', None):
    raise ValueError(f"The specified Array API version '{api_version}' is not supported.")
```
WDYT?

PR #93, opened 2024-02-22
Workaround np.linalg.solve ambiguity

asmeurer:
NumPy's solve() does not handle the ambiguity around x2 being 1-D vector vs. an n-D stack of matrices in the way that the standard specifies. Namely, x2 should be treated as a 1-D vector iff it is 1-dimensional, and a stack of matrices in all other cases. This workaround is borrowed from array-api-strict.

See https://github.com/numpy/numpy/issues/15349 and https://github.com/data-apis/array-api/issues/285.

Note that this workaround only works for NumPy. CuPy currently does not support stacked vectors for solve() (see
https://github.com/cupy/cupy/blob/main/cupy/cublas.py#L43), and the workaround in cupy.array_api.linalg does not seem to actually function.


asmeurer:
Test failure is an upstream issue https://github.com/data-apis/array-api-tests/issues/238

PR #99, opened 2024-03-05
Temporarily pin hypothesis on CI

asmeurer:
This is needed until https://github.com/data-apis/array-api-tests/issues/238 is fixed.


asmeurer:
We should probably update the minimum numpy version from 1.21 to something newer https://numpy.org/neps/nep-0029-deprecation_policy.html, but I'm not going to worry about that for now.

PR #103, opened 2024-03-09
HTML documentation

### LINKED ISSUE #44: Actual docs
Right now the only docs are the README, but as we add more compatibility functions, it may be better to have some actual documentation.

asmeurer:
Fixes #44


docs/dev/implementation-notes.md
```diff
@@ -0,0 +1,160 @@
...
+    re-export names from an external namespace.
+
+  - All wrapper functions pass `**kwargs` through to the wrapped function.
+
+  The onus is on users of array-api-compat to ensure they are only using
+  portable array API behavior, e.g., by testing against [array-api-strict](array-api-strict).
+
+
+## Avoiding Code Duplication
+
```
rgommers, line 74: This section may want to add a mention of JAX and Dask, even to just mention that they exist. Or use "for example" when namespaces are explicitly listed

docs/dev/implementation-notes.md
```diff
@@ -0,0 +1,160 @@
...
+The majority of the behavior for array-api-compat is tested by the
+[array-api-tests](https://github.com/data-apis/array-api-tests) test suite for
+the array API standard. There are also specific tests in
+[`tests/`](https://github.com/data-apis/array-api-compat/tree/main/tests).
+These tests should be limited to things that are not tested by the test suite,
+e.g., tests for [helper functions](../helper-functions.md) or for behavior that is
+not strictly required by the standard.
+
+array-api-tests is run against all supported libraries are tested on CI
+(except for JAX,
```
rgommers, line 151: the "except for JAX, wrapping" phrasing is a little odd - rephrase?

rgommers (approved):
This reads quite well, thanks Aaron. I didn't try to build locally, did read the whole diff and things are structured well. I only had a few minor comments.

It may be worth adding a CI job to run on PRs, failing on Sphinx warnings/errors, and then to deploy to GitHub Pages on merges. Can also be done as a follow-u. Either way, LGTM to merge this.

asmeurer:
I actually haven't finished here (hence the "draft" status), but thanks for taking a look.

I would especially like your thoughts on this section https://github.com/data-apis/array-api-compat/pull/103/files#diff-ebbf0983decf5c6b5520f86b136672b0a044a778687a8813de4900f4f554d58eR7. Specifically, this paragraph, https://github.com/data-apis/array-api-compat/pull/103/files#diff-ebbf0983decf5c6b5520f86b136672b0a044a778687a8813de4900f4f554d58eR36-R40. I think it would be good to have a policy on what we will and won't include in this library. A lot of people have made suggestions for things that could go here which I personally feel would be out of scope (e.g., I'm hesitant on even #104, because it's outside the domain of the standard, and I'm *really* hesitant on suggestions to include things like additional helpers for functions that aren't in the standard).

asmeurer:
Actually we should probably add a separate "Scope" section to the main part of the docs to make this really clear to everyone. But I want to make sure that we are in alignment about what is and isn't in scope for this package.

rgommers:
> Actually we should probably add a separate "Scope" section to the main part of the docs to make this really clear to everyone. But I want to make sure that we are in alignment about what is and isn't in scope for this package.

I agree with the way you have written it now. That matches my current understanding of what we're accepting into `array-api-compat`. Maybe the one thing to add is that functions accepted for inclusion in the standard, but not yet merged into a draft version of the standard or in a released version, may already be added to `array-api-compat`.

asmeurer:
I've added a section to the main page explaining the scope of the package https://github.com/data-apis/array-api-compat/pull/103/files#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6R124

asmeurer:
This is ready for review. You can see the rendered HTML version at https://data-apis.org/array-api-compat/

rgommers (approved):
> I've added a section to the main page explaining the scope of the package

This looks quite good, it captures the intent and rationale well. 

The main page is structured nicely too. I don't have substantial comments, it seems good to go. Nice work!

Issue #106, opened 2024-03-10
`array_api_compat.numpy.asarray(torch.Tensor)` returns a `torch.Tensor`


izaid:
This feels like a bug to me. My expectation is that `xp.asarray` converts to an array object in the `xp` namespace if that operation is supported.

```
>>> import torch
>>> import array_api_compat.numpy
>>> x = torch.arange(10)
>>> array_api_compat.numpy.asarray(x)
tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

rgommers:
Agreed, that is a bug. This line looks wrong:
https://github.com/data-apis/array-api-compat/blob/d10bf57e73e88f692c59b420eb0c928abbb3db1c/array_api_compat/common/_aliases.py#L352

ivirshup:
What is the expected behavior for `xp.asarray(some_other_array_type)`? Are there cases it should fail in?

asmeurer:
If the input type is one of the types mentioned in the standard (in the type signature), it should convert that input type. Otherwise, generally for array-api-compat, any behavior outside of the standard should just pass through to the library function. So in this case, asarray() should indeed convert the input because normal `np.asarray(Tensor)` works (and also I think `Tensor` falls under the category of `SupportsBufferProtocol`).

lithomas1:
Yeah, I messed up asarray. I have a WIP branch fixing that ,and some other dask stuff.

asmeurer:
OK, I'll hold off on looking at this for now so I don't step on your toes. You may want to integrate the test from https://github.com/data-apis/array-api-compat/pull/109/files. It may also be necessary to update the copy logic generally similar to https://github.com/data-apis/array-api-strict/pull/15. I don't know if the "check if a copy was made" trick can work for cupy and dask, but if it can we should use it. If there end up being too many library-specific checks in the `asarray` code it may be more prudent to just split it into separate helpers for each library.

lithomas1:
@izaid @ivirshup 

Are you able to test main again? I should have this bug fixed, but just wanted to make sure.

izaid:
> @izaid @ivirshup
> 
> Are you able to test main again? I should have this bug fixed, but just wanted to make sure.

Is there a link somewhere with steps to follow? I only have `array_api_compat` installed from pip.

lucascolley:
- clone `array_api_compat` main
- `cd array_api_compat`
- create a new env
- `pip install torch`
- then run your example above

[...]

PR #109, opened 2024-03-11
`asarray` partial fix

ivirshup:
Attempt at a fix for `asarray`.

* https://github.com/data-apis/array-api-compat/issues/106

Opening as a draft in case someone else gets the opportunity to look at this earlier, as some cases still error. I think the test should be useful to anyone else attempting this.

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -42,6 +42,17 @@ def test_device(library):
     x2 = to_device(x, dev)
     assert device(x) == device(x2)
 
+@pytest.mark.parametrize("target_library,func", is_functions.items())
+@pytest.mark.parametrize("source_library", is_functions.keys())
+def test_asarray(source_library, target_library, func):
+    src_lib = import_(source_library, wrapper=True)
+    tgt_lib = import_(target_library, wrapper=True)
+    is_tgt_type = globals()[func]
+
+    a = src_lib.asarray([1, 2, 3])
+    b = tgt_lib.asarray(a)
+
+    assert is_tgt_type(b), f"Expected {b} to be a {tgt_lib.ndarray}, but was {type(b)}"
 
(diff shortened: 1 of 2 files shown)


asmeurer:
I'm not sure what the right fix here is. The code was correct before the `__array__` check was added in 8ee16138094d44a65be73af4b88a5b3cc74d2203, but I'm not completely clear what the purpose of that check was. 

Also, this code seems to deal mostly with `copy` logic, and that logic needs to be refactored, similar to https://github.com/data-apis/array-api-strict/pull/15 (at the very least, copy=False should work in numpy 2.0). This version of the function is going to be more complicated than the one in array-api-strict though because it also has to support things outside of the standard (i.e., any valid input to `np.asarray`), and also because it's a wrapper function that's reused for numpy, cupy, and dask.

tests/test_common.py
```diff
@@ -42,6 +42,17 @@ def test_device(library):
     x2 = to_device(x, dev)
     assert device(x) == device(x2)
 
+@pytest.mark.parametrize("target_library,func", is_functions.items())
+@pytest.mark.parametrize("source_library", is_functions.keys())
+def test_asarray(source_library, target_library, func):
```
asmeurer, line 47: +1 to this test. We are going to need to add a lot of tests here for asarray, partly because the test suite only handles things within the scope of the standard, and partly because the test suite itself doesn't test as much as it could (https://github.com/data-apis/array-api-tests/issues/241). We can also adopt the tests from https://github.com/data-apis/array-api-strict/pull/15 (once we have fixed the `copy` behavior, but I expect that will need to just be part of this overall fix since this is part of that logic).

asmeurer:
This was included with https://github.com/data-apis/array-api-compat/pull/110

PR #115, opened 2024-03-18
Add a test that there are no hard dependencies

asmeurer:


asmeurer:
array-api-strict requires Python 3.9 but we are still testing against 3.8. This isn't *that* important. I'll just leave this here for now.

PR #118, opened 2024-03-20
Only run the publish to pypi build on tags

asmeurer:
Otherwise this build "fails" on main leading to a bunch of annoying failed actions checks.


asmeurer:
Might as well merge this before a release to see if it works.

asmeurer:
It no longer fails on main so that's good.

PR #119, opened 2024-03-20
Support the copy keyword in asarray

asmeurer:
Not finished yet. Still need to:

- [ ] Make sure everything is tested on CI, including NumPy 1.21, 1.26, and 2.0
- [x] Implement CuPy and dask.array support. 
- [x] Fix the device keyword in cupy.asarray (currently it is ignored!)


array_api_compat/cupy/_aliases.py
```diff
@@ -62,6 +61,42 @@
...
+    specification for more details.
+
+    'namespace' may be an array module namespace. This is needed to support
+    conversion of sequences of Python scalars.
+    """
+    with cp.cuda.Device(device):
+        # cupy is like NumPy 1.26 (except without _CopyMode). See the comments
+        # in asarray in numpy/_aliases.py.
+        if copy is None:
+            copy = False
```
asmeurer, line 94: @leofang did you say at the meeting today that cupy 14 would definitely support NumPy 2.0? The meaning of copy=False changed between NumPy 1.26 and 2.0. Is it safe to add a `if cupy.__version__ >= 14` check here? If we just leave this wrapper code as it is, it will break once cupy makes the change, because the default would become the array API copy=False semantics.

leofang: It was TBD so here's my question: I would think like dpctl once CuPy is compliant we'd no longer need array-api-compat coverage? If so, I don't think we need a version check here, as it'd become moot anyway? WDYT?

asmeurer: The way the compat library currently works is unconditionally wraps cupy. We could remove the wrapping once cupy is compliant but we'd need a version check to do that too.

leofang: I mean, the version check could be added to, say, `__init__.py` instead of here, once we know better after which CuPy version all array-api-compat wrappers can just fall back to native CuPy without extra action (=no-ops), right? Maybe I oversimplified it?

asmeurer: If cupy is completely compliant and needs no wrappers we could do that. We're not doing that yet for numpy 2.0 but we probably should. 

At any rate, I've thought of a better way to structure this code so that it won't actually break when cupy changes the meaning of `copy=False`, so this isn't actually important. I'll still need to update this to remove the `NotImplementedError` for `copy=False`, but is isn't pressing to try to do it ahead of time now because I've made it so that the default behavior won't break, and that would be a new feature for cupy anyways.

asmeurer:
@lithomas1 can you take a look at the dask.array changes here.

asmeurer:
Python 3.8 is starting to become a nuisance. I think I'm going to just drop support for it.

lithomas1:
At an initial glance, the dask changes look fine to me.

CI seems to be failing though.

(looks like something is forcing an older jax on Python 3.10)

PR #126, opened 2024-03-27
Don't wrap NumPy 2.0 at all

asmeurer:
This also adds a `use_compat` flag to `array_namespace`. If None, it does the default behavior. If True or False it forces the use or
non-use of the compat wrapper.

Tests will fail here until https://github.com/data-apis/array-api-strict/pull/24 is released.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -178,7 +178,7 @@ def _check_api_version(api_version):
     elif api_version is not None and api_version != '2022.12':
         raise ValueError("Only the 2022.12 version of the array API specification is currently supported")
 
-def array_namespace(*xs, api_version=None, _use_compat=True):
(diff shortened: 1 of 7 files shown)


asmeurer:
Note that I'm still running the test suite against numpy 2.0, as I want to make sure that if at any point in the future wrapping becomes necessary again that I am made aware of that.

asmeurer:
This is ready to go. 

Once we start supporting 2023.12, we'll probably have to add a wrapper again, if just for the new inspection APIs. My guess is that if/when that happens we'll want to have a separate wrapper namespace for NumPy 2.0 because the rest of the NumPy 1.26 wrapper functions are pointless to use with 2.0. Maybe at that point we can reuse whatever solution I come up with for https://github.com/data-apis/array-api-strict/issues/8 to streamline this.

PR #130, opened 2024-04-15
Ensure that Jax float0 array is recognized

NeilGirdhar:
See https://github.com/google/jax/issues/20620.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -18,6 +18,20 @@ import math
 import inspect
 import warnings
 
+def _is_jax_zero_gradient_array(x):
+    """Return True if `x` is a zero-gradient array.
+
+    These arrays are a design quirk of Jax that may one day be removed.
+    See https://github.com/google/jax/issues/20620.
+    """
+    if 'numpy' not in sys.modules or 'jax' not in sys.modules:
+        return False
+
+    import numpy as np
+    import jax
+
+    return isinstance(x, np.ndarray) and x.dtype == jax.float0
+
 def is_numpy_array(x):
     """
     Return True if `x` is a NumPy array.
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
This is intended to be a temporary solution until Jax (ideally) makes these arrays into Jax arrays.

NeilGirdhar:
(Error appears to be unrelated to this pull request.)

asmeurer:
This looks good. Can you add a test.

NeilGirdhar:
Great, I added a test.  If you want, I could also add:
```python
assert isinstance(jax_zero, np.ndarray)  # If this triggers, the workaround can probably be removed
```

array_api_compat/common/_helpers.py
```diff
@@ -18,6 +18,20 @@
 import inspect
 import warnings
 
+def is_jax_zero_gradient_array(x):
```
asmeurer, line 21:
```suggestion
def _is_jax_zero_gradient_array(x):
```

array_api_compat/common/_helpers.py
```diff
@@ -44,7 +58,8 @@ def is_numpy_array(x):
     import numpy as np
 
     # TODO: Should we reject ndarray subclasses?
-    return isinstance(x, (np.ndarray, np.generic))
+    return (isinstance(x, (np.ndarray, np.generic))
+            and not is_jax_zero_gradient_array(x))
```
asmeurer, line 62:
```suggestion
            and not _is_jax_zero_gradient_array(x))
```

array_api_compat/common/_helpers.py
```diff
@@ -149,7 +164,7 @@ def is_jax_array(x):
 
     import jax
 
-    return isinstance(x, jax.Array)
+    return isinstance(x, jax.Array) or is_jax_zero_gradient_array(x)
```
asmeurer, line 167:
```suggestion
    return isinstance(x, jax.Array) or _is_jax_zero_gradient_array(x)
```

asmeurer:
The tests failure should be fixed by my suggestion, and the test suite failure should be fixed by https://github.com/data-apis/array-api-compat/pull/131.

NeilGirdhar:
(Done.)

PR #133, opened 2024-04-23
Add det/slogdet workaround for dask

lithomas1:


asmeurer:
Is this needed for a downstream package?

This is right on the border of something that's too algorithmically complex to be appropriate for array-api-compat https://data-apis.org/array-api-compat/#scope. I'm not saying I'm rejecting it outright, but it does worry me. Note that the array API test suite does not have any sort of numerical test for these functions. It only tests basic things like that the output shapes and dtypes are correct. At the very least I would start with taking existing tests from other libraries and seeing if they can be adapted for these functions. 

Another thing to consider here is whether this can be upstreamed to Dask itself.

PR #134, opened 2024-05-03
WIP: Add `sparse` compatibility layer.

hameerabbasi:
This PR adds a compatibility layer for `sparse`.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -50,6 +50,7 @@ def is_numpy_array(x):
     is_torch_array
     is_dask_array
     is_jax_array
+    is_sparse_array
     """
     # Avoid importing NumPy if it isn't already
     if 'numpy' not in sys.modules:
@@ -79,6 +80,7 @@ def is_cupy_array(x):
     is_torch_array
     is_dask_array
     is_jax_array
+    is_sparse_array
     """
     # Avoid importing NumPy if it isn't already
     if 'cupy' not in sys.modules:
@@ -105,6 +107,7 @@ def is_torch_array(x):
     is_cupy_array
     is_dask_array
     is_jax_array
+    is_sparse_array
     """
     # Avoid importing torch if it isn't already
     if 'torch' not in sys.modules:
@@ -131,6 +134,7 @@ def is_dask_array(x):
     is_cupy_array
     is_torch_array
     is_jax_array
+    is_sparse_array
     """
     # Avoid importing dask if it isn't already
     if 'dask.array' not in sys.modules:
@@ -157,6 +161,7 @@ def is_jax_array(x):
     is_cupy_array
     is_torch_array
     is_dask_array
+    is_sparse_array
     """
     # Avoid importing jax if it isn't already
     if 'jax' not in sys.modules:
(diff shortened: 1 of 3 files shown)


array_api_compat/common/_helpers.py
```diff
@@ -166,6 +171,35 @@ def is_jax_array(x):
 
     return isinstance(x, jax.Array) or _is_jax_zero_gradient_array(x)
 
+
+def is_sparse_array(x) -> bool:
```
asmeurer, line 175: Is the package still called pydata/sparse? I worry this function name could be confusing out of context.

array_api_compat/sparse/__init__.py
```diff
@@ -0,0 +1,5 @@
+from sparse import * # noqa: F403
+from ..common._aliases import * # noqa: F403
```
asmeurer, line 2: This is not right. The aliases in common._aliases need to be wrapped in sparse/_aliases.py to work. See https://data-apis.org/array-api-compat/dev/implementation-notes.html

Furthermore, you should only wrap the functions that actually need wrapping. 

If you prefer, you can put all the compatibility shims for sparse in sparse itself, and keep the support here just limited to the helper functions. This is how it currently works for JAX. Since you're also the maintainer of sparse, I would suggest this approach.

array_api_compat/common/_helpers.py
hameerabbasi: I've taken to just calling it `sparse` in the past year to 18 months -- PyData/Sparse is a mouthful to say and difficult to type right.

array_api_compat/sparse/__init__.py
hameerabbasi: Thanks for the insight -- fixed in https://github.com/data-apis/array-api-compat/pull/134/commits/a4966e7f64309581e3003985f72bef6ce2787a8b.

array_api_compat/common/_helpers.py
hameerabbasi: I've reworded the docstring in https://github.com/data-apis/array-api-compat/pull/134/commits/fb14cc815b9cb0e4b7d2ae8180b685537443c827 for clarity, and marking as resolved -- however, please mark as un-resolved and propose a better name if you can think of one.

hameerabbasi:
Given that JAX also doesn't have CI and we already have CI for the Array API in `sparse`, I feel this is complete.

array_api_compat/common/_helpers.py
asmeurer: I'm curious what other people think of this. @rgommers

My worry is that `is_sparse_array` will be a confusing name, since it is not checking ia an array is sparse, but rather if it comes from the *library* "sparse". But this also matches the convention of the other functions in this namespace.

asmeurer:
Can you update the docs. If you search for the places where JAX is mentioned you can also mention sparse there in the same way.

asmeurer:
Also sparse can be added to the same the places where jax is tested in the tests.

hameerabbasi:
> Can you update the docs. If you search for the places where JAX is mentioned you can also mention sparse there in the same way.

Done in https://github.com/data-apis/array-api-compat/pull/134/commits/d3c6636bdcf878b98711373c9fab99b74b924f6d

> Also sparse can be added to the same the places where jax is tested in the tests.

Done in https://github.com/data-apis/array-api-compat/pull/134/commits/855756d17727bb2998c846a7c9a230131a525e46

array_api_compat/common/_helpers.py
hameerabbasi: Done in https://github.com/data-apis/array-api-compat/pull/134/commits/7ebc3c096ad022f1b3448ca233377a3ec72b3af8.

hameerabbasi:
CI seems to [fail](https://github.com/data-apis/array-api-compat/actions/runs/9090494121/job/24983380893?pr=134#step:5:277) with:

```
E           ImportError: Numba needs NumPy 1.22 or greater. Got NumPy 1.21.
```

I don't see how we can get around this; Numba is one of the requirements of `sparse`.

hameerabbasi:
The PyTorch failures seem to be unrelated to this PR, from a quick glance.

asmeurer:
This page should be updated too https://github.com/data-apis/array-api-compat/blob/main/docs/supported-array-libraries.md

hameerabbasi:
> This page should be updated too https://github.com/data-apis/array-api-compat/blob/main/docs/supported-array-libraries.md

Done in https://github.com/data-apis/array-api-compat/pull/134/commits/b77472d1c9ec9c32b6932c7f2153da2f096071e0.

asmeurer:
OK, this looks good. Let me know if you need this released soon. Otherwise, it might take a bit because I need to figure out what is going on with the PyTorch failures.

Issue #136, opened 2024-05-09
`sign` special case implementations


mdhaber:
According to v2022.12 (and v2023.12) of the array API standard, the special cases of `sign` include:

> For real-valued operands... If x_i is NaN, the result is NaN.

However,

```python3
from array_api_compat import numpy as np, cupy as cp, torch
np.sign(np.asarray(np.nan))  # nan
cp.sign(cp.asarray(cp.nan))  # array( 0.000e+00)
torch.sign(torch.asarray(torch.nan))  # tensor(0.)
```

There may be other special cases that are not yet implemented. I haven't done a complete review, but I noticed that `torch` gives an error when the input is complex.
```python3
torch.sign(torch.asarray(1+1j))
# RuntimeError: Unlike NumPy, torch.sign is not intended to support complex numbers. Please use torch.sgn instead.
```

asmeurer:
I haven't tried to do any special-cases workarounds here yet. I guess if this is causing issues for you we can add a workaround.

asmeurer:
We definitely should add complex support to torch.sign. It sounds like that will just be an easy wrapper of torch.sgn.

mdhaber:
Yeah I actually did run into that special case while converting code from NumPy to array API. The purpose of the code was to produce a different status flag that depends on whether an element is positive, negative, zero, or NaN, so no wonder I ran into it.

asmeurer:
We can add it. The main concern with adding special-case handling is it means adding a mask to the function, so it could be a minor hit to performance. I would at least make sure there are upstream issues about this to the libraries that don't implement it correctly.

rgommers:
> The main concern with adding special-case handling is it means adding a mask to the function, so it could be a minor hit to performance. 

I think that won't be a minor hit. For element-wise functions, adding `isnan` checks and masking may slow things down by 2x or more.

In general, these special cases have not been well validated yet, so I'd be quite reluctant to assume they are all correctly specified or support them in the compat layer.

In this case, it may be possible to fix in NumPy. There's a bunch of special-casing to make `sign(nan)` return `nan`, which looks like old code and disagrees with C99's `signbit` on purpose:

https://github.com/numpy/numpy/blob/2a9b9134270371b43223fc848b753fceab96b4a5/numpy/_core/src/umath/loops.c.src#L1341-L1350

[...]

PR #137, opened 2024-05-10
Fix sign() for torch and cupy

asmeurer:
Neither propagate nans correctly, and torch does not support complex numbers.

Fixes https://github.com/data-apis/array-api-compat/issues/136


asmeurer:
https://github.com/data-apis/array-api-compat/issues/136 should be resolved before this is merged, specifically, we should decide if it's worth fixing the sign(nan) special case, and if we want to keep that special case at all. Regardless of that, though, we should keep the torch complex handling, as it's very straightforward to implement.

asmeurer:
It seems that torch has gained quite a few new test failures since the last time we ran them. I don't know if that's because of a test suite update or a torch update.

asmeurer:
It sounds like this will be useful, so I'm going to merge.

Issue #150, opened 2024-06-17
Add ndonnx


adityagoel4512:
[ndonnx](https://github.com/Quantco/ndonnx) is an ONNX-backed array library. It implements the Array API standard and support is actually directly in the main namespace. ndonnx enables users to export Array API compliant code to [ONNX](https://github.com/onnx/onnx) automatically, a process which traditionally required quite a lot of code duplication.

Since we don't need an array-api-compat wrapper for ndonnx, adding it here would be somewhat similar to the [JAX approach](https://github.com/data-apis/array-api-compat/pull/84) - we also run array-api-tests over at ndonnx itself.

Would it be acceptable to add ndonnx here?

asmeurer:
If a package is full array API compatible, it doesn't need to add anything here. It will just work out of the box. As long as you define `__array_namespace__` correctly, the `array_namespace()` helper will use it when passed an ndonnx array. 

The only code in array-api-compat for libraries like JAX or NumPy 2.0 that are fully array API compatible is helper functions like `is_numpy_array`, which are needed for downstream code that needs to write NumPy-specific paths for things like performance or to workaround some limitations of the array API. See https://data-apis.org/array-api-compat/helper-functions.html.

If you anticipate packages needing an `is_ndonnx_array` helper, we can add that. 

Otherwise, there's nothing really to add here. 

We could perhaps add a page to the documentation listing other array API compatible packages like ndonnx that aren't explicitly part of array-api-compat.

adityagoel4512:
> If you anticipate packages needing an is_ndonnx_array helper, we can add that.

I think this main utility. In particular, for ONNX export you cannot use any implicitly eager constructs part of the Array API standard (like `__float__()`) since there is no available data to collect and instead we raise an exception. In such a place you would branch and provide an ONNX specific code path.

asmeurer:
You can make a PR for this if you like.

Although it sounds like this is a special case of the behavior that was specified at https://github.com/data-apis/array-api/pull/652. I wonder if we shouldn't try to make something more general. I'm also wondering if the proper way for libraries to handle this should be to catch `ValueError` and branch based on that, given that's now what the standard specifically says should happen. 

[...]

PR #154, opened 2024-06-24
Add ndonnx

adityagoel4512:
Closes https://github.com/data-apis/array-api-compat/issues/150.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,6 +28,11 @@ jobs:
           else
             PIP_EXTRA='numpy==1.26.*'
           fi
+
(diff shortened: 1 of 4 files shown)


adityagoel4512:
Closed this since the CI tests as far back as python 3.9, which ndonnx dropped support for in line with https://scientific-python.org/specs/spec-0000/. Changing the minimally supported python version is beyond the scope of this PR.

asmeurer:
That's just there to make sure older versions of NumPy work. You should be able to add it to the skips in the CI file.

adityagoel4512:
Does the NumPy skips file need to be updated since the release of NumPy 2.0?

asmeurer:
Yes, you can ignore that. I'm fixing it in another PR.

PR #155, opened 2024-06-25
Update NumPy CI for NumPy 2.0

asmeurer:


asmeurer:
The publish distributions thing is an upstream issue with twine that will hopefully be fixed before we want to do an actual release.

Issue #164, opened 2024-07-27
BUG: `array_namespace` returns compat for NumPy scalars with `use_compat=None`


lucascolley:
With array-api-compat 1.8, NumPy 2.0.1

```shell
In [22]: array_namespace(np.asarray(0.0), use_compat=None)
Out[22]: <module 'numpy' from '/Users/lucascolley/mambaforge/envs/scipy-dev/lib/python3.10/site-packages/numpy/__init__.py'>

In [23]: array_namespace(np.float64(0.0), use_compat=None)
Out[23]: <module 'scipy._lib.array_api_compat.numpy' from '/Users/lucascolley/programming/scipy/build-install/lib/python3.10/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>
```

This behaviour is quite problematic for SciPy given NumPy's tendency to throw NumPy scalars at you under very little pressure. 

```shell
In [25]: array_namespace(np.asarray(0.0) - 32, use_compat=None)
Out[25]: <module 'scipy._lib.array_api_compat.numpy' from '/Users/lucascolley/programming/scipy/build-install/lib/python3.10/site-packages/scipy/_lib/array_api_compat/numpy/__init__.py'>
```

It seems practical to have NumPy scalars return the same namespace as NumPy arrays when `use_compat=None`. Or perhaps a different argument should have to be passed to `use_compat` (that would be fine for us).

lucascolley:
discovered in https://github.com/scipy/scipy/pull/21264#issuecomment-2254097094

rgommers:
> It seems practical to have NumPy scalars return the same namespace as NumPy arrays when `use_compat=None`.

Agreed, that should be the right default.

lucascolley:
fix at gh-165

asmeurer:
Seems like NumPy should be fixed here too.

rgommers:
Can you spell out what should change in NumPy? It doesn't have an `array_namespace` function, so I don't immediately see it.

lucascolley:
Do you mean giving NumPy scalars a matching `__array_namespace__` attribute?

asmeurer:
Yes. This should work

```py
>>> np.asarray(0).__array_namespace__()
<module 'numpy' from '/Users/aaronmeurer/miniconda3/envs/numpy2/lib/python3.12/site-packages/numpy/__init__.py'>
>>> np.asarray([0])[0].__array_namespace__()
Traceback (most recent call last):
[...]

PR #165, opened 2024-07-27
BUG: fix `array_namespace` for NumPy scalars

lucascolley:
closes gh-164


tests/test_array_namespace.py
```diff
@@ -36,6 +36,13 @@ def test_array_namespace(library, api_version, use_compat):
         else:
             assert namespace == getattr(array_api_compat, library)
 
+    if library == "numpy":
+        # check that the same library is returned for NumPy scalars
```
lucascolley, line 40:
```suggestion
        # check that the same namespace is returned for NumPy scalars
```

Issue #152, opened 2024-06-17
Result of `numpy.sum` with `float32` input is `float64`


mdhaber:
IIUC the standard states that input dtype is to be preserved for floating point types when `dtype` is unspecified.

> **dtype** - data type of the returned array. If `None`, the returned array must have the same data type as `x`, unless `x` has an integer data type

However,
```python3
from array_api_compat import numpy as np
x = np.arange(10, dtype=np.float32)
np.sum(x).dtype  # dtype('float64')
```

asmeurer:
This is part of #127. This was changed in 2023.12. I don't see a need to keep the previously standardized behavior intact. Unfortunately, I haven't yet started work on updating array-api-compat for 2023.12, but I plan to start that work now.

mdhaber:
Interesting. I didn't notice that the old standard required that the output `dtype` be the default floating point type. Glad that was changed!

Issue #127, opened 2024-03-28
Support for 2023.12


asmeurer:
This is a tracking issue for 2023.12 support in array-api-compat.

The list of changes for 2023.12 is at https://data-apis.org/array-api/latest/changelog.html#v2023-12

Currently, we don't have test suite support for 2023.12 (https://github.com/data-apis/array-api-tests/issues/249), so it's hard to implement full support. However, if there are any known differences for a given library and 2023.12, we can implement a wrapper for them, especially if this is blocking anyone's real world usage. Please comment below if this is the case for you. 

Note that any function that is already implemented in upstream libraries will already work with array-api-compat, since it only extends the existing library namespaces. So this is only necessary if some upstream function or behavior is different from what is written in the 2023.12 standard. 

Here is the checklist of functions (from https://github.com/data-apis/array-api-tests/issues/249):

Here is the list of changes https://data-apis.org/array-api/latest/changelog.html#v2023-12

**New Functions**

- [x] `__array_namespace_info__` (i.e., the [inspection API](https://data-apis.org/array-api/latest/API_specification/inspection.html))
- [x] `clip` (not implemented for dask, as the wrapper implementation uses mask assignment)
- [x] `copysign`
- [x] `cumulative_sum`
- [x] `hypot`
- [x] `maximum`
- [x] `minimum`
- [x] `moveaxis`
- [x] `repeat`
- [ ] `searchsorted`
- [x] `signbit`
- [ ] `tile`
- [x] `unstack`

**API Changes**

- [ ] `device` keyword added to `astype`
- [ ] ~`from_dlpack` requires exceptions in some cases~ (won't be wrapped)
- [ ] ~`copy` and `device` keywords to `from_dlpack`~ (won't be wrapped)
- [x] `sum`, `prod`, `trace` output dtype change (https://github.com/data-apis/array-api-tests/issues/234)

**Other Changes**

- [ ] `tensordot` should support negative axes

asmeurer:
https://github.com/data-apis/array-api-compat/pull/166

asmeurer:
@leofang I'm looking at implementing the inspection APIs for CuPy https://data-apis.org/array-api/latest/API_specification/inspection.html. I have some questions:

- How can I get a list of all devices (for `devices()`). Is this correct?
  ```py
  def devices():
      return [cupy.Device(i) for i in range(cupy.cuda.runtime.getDeviceCount())]
  ```
- Is `cupy.Device(0)` always the default device?
[...]

PR #166, opened 2024-07-29
More fixes for 2023.12 support

asmeurer:
#127 

Fixes https://github.com/data-apis/array-api-compat/issues/152


array_api_compat/torch/_info.py
asmeurer, line 1: I'm hoping someone with more knowledge of pytorch can review this. I've made various assumptions here, which I'll try to outline in comments below. If anyone can confirm whether those assumptions are valid, that would be helpful.

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        --------
+        >>> info = np.__array_namespace_info__()
+        >>> info.capabilities()
+        {'boolean indexing': True,
+         'data-dependent shapes': True}
+
+        """
+        return {
+            "boolean indexing": True,
+            "data-dependent shapes": True,
```
asmeurer, line 87: I'm assuming boolean indexing and data-dependent shapes (i.e., functions like `unique`) always work in pytorch.

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+            The default device used for new PyTorch arrays.
+
+        Examples
+        --------
+        >>> info = np.__array_namespace_info__()
+        >>> info.default_device()
+        'cpu'
+
+        """
+        return torch.device("cpu")
```
asmeurer, line 115: I'm assuming the default device in pytorch is always "cpu" (note there is currently an unfortunate distinction between the "default device" and "current device" in the standard. See https://github.com/data-apis/array-api/issues/835). By "default device", I mean the device that is used by default when pytorch is first started.

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        --------
+        >>> info = np.__array_namespace_info__()
+        >>> info.default_dtypes()
+        {'real floating': torch.float32,
+         'complex floating': torch.complex64,
+         'integral': torch.int64,
+         'indexing': torch.int64}
+
+        """
+        default_floating = torch.get_default_dtype()
```
asmeurer, line 150: If I'm understanding the docs correctly, this function will always give the default floating-point dtype (i.e., what is used by default for something like `torch.ones()` with no dtype argument).

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        default_complex = torch.complex64 if default_floating == torch.float32 else torch.complex128
```
asmeurer, line 151: The docs explicitly state that the default complex dtype always matches the floating dtype. https://pytorch.org/docs/stable/generated/torch.set_default_dtype.html

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        default_integral = torch.asarray(0, device=device).dtype
```
asmeurer, line 152: Is there a way to access this that doesn't require creating a tensor?

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        return {
+            "real floating": default_floating,
+            "complex floating": default_complex,
+            "integral": default_integral,
+            "indexing": default_integral,
```
asmeurer, line 157: I'm assuming the default indexing type is always the same as the default integer dtype (the default indexing type is a concept in the array API for functions that return indexing arrays. For example, `nonzero` should return an array with the default indexing type https://data-apis.org/array-api/latest/API_specification/generated/array_api.nonzero.html)

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+    def _dtypes(self, kind):
+        bool = torch.bool
+        int8 = torch.int8
+        int16 = torch.int16
+        int32 = torch.int32
+        int64 = torch.int64
+        uint8 = getattr(torch, "uint8", None)
+        uint16 = getattr(torch, "uint16", None)
+        uint32 = getattr(torch, "uint32", None)
+        uint64 = getattr(torch, "uint64", None)
```
asmeurer, line 170: Are these the only dtypes that can be undefined (I know newer torch versions have them, but I want to make sure older versions work here too).

asmeurer: Also I know the support for some of these is limited. Would it be more correct to always omit them, even when they are technically defined here? They aren't really fully supported from the point of view of the array API standard.

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+         'int64': numpy.int64}
+
+        """
+        res = self._dtypes(kind)
+        for k, v in res.copy().items():
+            if v is None:
+                del res[k]
+                continue
+            try:
+                torch.empty((0,), dtype=v, device=device)
```
asmeurer, line 312: Is this the best way to test if a dtype is supported on a given device?

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+                "complex128": complex128,
+            }
+        if isinstance(kind, tuple):
+            res = {}
+            for k in kind:
+                res.update(self.dtypes(kind=k))
+            return res
+        raise ValueError(f"unsupported kind: {kind!r}")
+
+    @cache
```
asmeurer, line 251: I'm assuming it's safe to cache the output of this function (it's potentially expensive since it constructs tensors on the given device to test if a dtype is supported).

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+            if v is None:
+                del res[k]
+                continue
+            try:
+                torch.empty((0,), dtype=v, device=device)
+            except:
+                del res[k]
+        return res
+
+    @cache
```
asmeurer, line 317: I'm assuming the set of legal devices never changes at runtime and can be cached.

array_api_compat/torch/_info.py
```diff
@@ -0,0 +1,369 @@
...
+        >>> info = np.__array_namespace_info__()
+        >>> info.devices()
+        [device(type='cpu'), device(type='mps', index=0), device(type='meta')]
+
+        """
+        # Torch doesn't have a straightforward way to get the list of all
+        # currently supported devices. To do this, we first parse the error
+        # message of torch.device to get the list of all possible types of
+        # device:
+        try:
```
asmeurer, line 345: This is the big one. This is code is based on some discussions with @pearu. To get the list of possible torch devices, we first parse an error message, then check which of those devices actually work. If there is any better way of doing this, please let me know. I would definitely prefer if this functionality were built-in to pytorch (ditto for the other functions here too).

amjames: This is unfortunate. I think there might be a less expensive, but more ugly way to do this. Would adding something to surface this info even help, or would you still need to have something like this to support older versions?

array_api_compat/torch/_info.py
amjames: What defines supported? 

If dtype and device are valid this should always work, storage is untyped so the dtype is only used to compute the number of bytes needed. 

Any given operator may or may not correctly dispatch for the underlying dtype. If the type is not explicitly handed for a given operator it will throw.

array_api_compat/torch/_info.py
amjames: Yes, this is true. However that "default" in torch is not static. It can be changed by the user, so that would be the "current default" and the "default default" would be `torch.float32`

array_api_compat/torch/_info.py
amjames: You can also just hard code it, we don't have a default for integers internally. If this were to change it would break bc and would need to be updated everywhere an integer tensor is produced, manually.

array_api_compat/torch/_info.py
amjames: I would say that depends on the purpose of the declaration. If it is to list the definitions for data types provided by the library leave them. If the point is to declare the data types supported by the array api then drop them. 

The two `float8` options are also going to have limited support, and possibly be undefined if the torch version is too old.

array_api_compat/torch/_info.py
amjames: There is currently no way to define a new dtype than adding an entry to a enum in the source code.

array_api_compat/torch/_info.py
asmeurer: Adding something would definitely be helpful. Note that ideally, pytorch will implement this exact API here, since it's [part of the array API](https://data-apis.org/array-api/latest/API_specification/inspection.html).

array_api_compat/torch/_info.py
asmeurer: You mean it's always `int64`? I wasn't sure if this would be different for certain devices.

array_api_compat/torch/_info.py
asmeurer: Just a note to myself: I checked and torch does correctly fail if you set the default dtype to float64 and try to create a tensor on a device that doesn't support float64:

```py
>>> torch.set_default_dtype(torch.float64)
>>> torch.asarray(0., device='mps')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64. Please use float32 instead.
```

(unlike certain other libraries that silently map float64 back to float32)

asmeurer: > so that would be the "current default" and the "default default" would be torch.float32

I'm not sure which this should use then. I *think* it should be the "current default", but the meaning of "default" is ambiguous. I mentioned this at https://github.com/data-apis/array-api/issues/835#issuecomment-2345183459

array_api_compat/torch/_info.py
asmeurer: I noticed that pytorch sort of has APIs to do this better, e.g., https://pytorch.org/docs/stable/generated/torch.mps.device_count.html#torch.mps.device_count, but they are not consistent across all device types, and I didn't want to hard-code all the possible device types here since torch seems to support a lot of them.

array_api_compat/torch/_info.py
asmeurer: I hadn't even considered the torch compiler. But I think as long as it functions without an error, that should be considered working (if that's not the case, then it might actually be worth flipping this flag in a compiler context, assuming that's easy). 

These flags exist for libraries like dask or JAX that straight up don't allow these types of operations.

array_api_compat/cupy/_info.py
```diff
@@ -0,0 +1,326 @@
...
+        --------
+        >>> info = xp.__array_namespace_info__()
+        >>> info.dtypes(kind='signed integer')
+        {'int8': cupy.int8,
+         'int16': cupy.int16,
+         'int32': cupy.int32,
+         'int64': cupy.int64}
+
+        """
+        # TODO: Does this depend on device?
```
asmeurer, line 233: @leofang is it possible for cupy to not support some of the array API dtypes depending on a given device?

array_api_compat/torch/_info.py
amjames: In that case I would say you are probably correct the way you have it. Those are intended to be supported and lack of support/conditional support in the compile context is considered a deficiency.

array_api_compat/torch/_info.py
amjames: We don't have a concept of the default integer type. It would always need to be provided or the type is deduced from the argument that is what happens in this case, python ints can be larger than int32 so we use int64.

I mean that I do not know of any mechanism for changing the behavior you see here. Our deduction rules for creation from a non array type can be seen [here](https://github.com/pytorch/pytorch/blob/63d6cd351a5425bd74bf45929da34254cf5bc040/torch/csrc/utils/tensor_new.cpp#L123). If the device does not support int64 I would expect you to see an error like you do with MPS and float64 (which I did not know would happen, I expect they had to do some work to make sure that error gets raised.)

amjames: Note that that layer of code is going to run "above" the dispatch to device specific logic (as in closer to python), so there is no way for the device to influence how that type is deduced, the device will be involved when the underlying tensor object is constructed where it will only see that a `dtype` argument has been provided.

array_api_compat/torch/_info.py
amjames: Note: I think that API should exist for optional devices. And you can find the module programmatically with `torch.get_device_module(<name>)`. The always available devices are of course special, but I can look into handling this on the pytorch side.

asmeurer:
So a couple of small things are still missing here https://github.com/data-apis/array-api-compat/issues/127. Most notable is `repeat` which is missing for PyTorch and has some minor issues with NumPy. There's also some concerns with `default_device` in the inspection API which might end up changing (see https://github.com/data-apis/array-api/issues/835). But there are quite a few changes here so I want to get them merged and released. I will hold off on bumping the `__array_api_version__` for now, though.

Issue #167, opened 2024-07-29
NumPy dev test failures


asmeurer:
Several tests for NumPy dev fail. What's concerning is that these don't seem to show up on CI, despite the CI apparently running the latest version of the test suite and installing the latest version of NumPy. See for instance https://github.com/data-apis/array-api-compat/actions/runs/10149966468/job/28066028900, which is definitely installing the right version of the test suite https://github.com/data-apis/array-api-compat/actions/runs/10149966468/job/28066028900#step:3:101, and definitely testing against numpy 2.1.0-dev https://github.com/data-apis/array-api-compat/actions/runs/10149966468/job/28066028900#step:7:24

```
FAILED array_api_tests/test_operators_and_elementwise_functions.py::test_clip - AssertionError: out.dtype=int64, but should be uint32 [clip(uint32)]
Falsifying example: test_clip(
    x=array([65535, 35892, 35892, 35892, 35892], dtype=uint32),
    data=data(...),
)
Draw 1 (min.shape, max.shape): ((), ())
Draw 2 (min): None
Draw 3 (max): array(0, dtype=int8)
Draw 4 (kwargs): {'max': array(0, dtype=int8)}
FAILED array_api_tests/test_set_functions.py::test_unique_inverse - TypeError: Input should be a NumPy array. It is a <class 'numpy.bool'> instead.
Falsifying example: test_unique_inverse(
    x=array(0., dtype=float32),
)
Explanation:
    These lines were always and only run by failing examples:
        /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/dtype_helpers.py:187
        /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/hypothesis_helpers.py:38
        /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/hypothesis_helpers.py:45
        /Users/aaronmeurer/Documents/numpy/numpy/_core/getlimits.py:491
FAILED array_api_tests/test_set_functions.py::test_unique_values - TypeError: Input should be a NumPy array. It is a <class 'numpy.bool'> instead.
Falsifying example: test_unique_values(
    x=array(0., dtype=float32),
)
Explanation:
    These lines were always and only run by failing examples:
        /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/dtype_helpers.py:187
        /Users/aaronmeurer/Documents/array-api-tests/array_api_tests/hypothesis_helpers.py:38
[...]

PR #170, opened 2024-07-30
Re-enable wrapping unconditionally for NumPy

### LINKED ISSUE #168: add `array_namespace` option to return wrapped if possible, else native
Since `array_namespace(x, use_compat=None)` now returns plain `numpy`, but released NumPy still has incompatibilities with the standard, we want to use `use_compat=True` in SciPy. However, since array-api-strict isn't wrapped, `array_namespace` throws a `ValueError`, so we have to use:

```python
try:
    return array_api_compat.array_namespace(*_arrays, use_compat=True)
except ValueError:
    return array_api_compat.array_namespace(*_arrays, use_compat=None)
```

array-api-strict isn't special here, this goes for any standard-compliant library. It would be nice to have an option to do this without having to catch an exception. Alternatively, we could just wait until NumPy is fully compatible with the standard before making the native namespace the default, but I'm not sure how feasible that is as the standard continues to change.

x-ref https://github.com/scipy/scipy/pull/21264

asmeurer:
NumPy 2.0 and 2.1 both have some incompatibilities with the array API (see
#167 and #164).

Closes #168.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,7 +6,7 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12']
-        numpy-version: ['1.21', '1.26', 'dev']
+        numpy-version: ['1.21', '1.26', '2.0', 'dev']
         exclude:
           - python-version: '3.11'
             numpy-version: '1.21'
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -145,7 +145,7 @@ def is_ndonnx_array(x):
 
     import ndonnx as ndx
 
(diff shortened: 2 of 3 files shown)


asmeurer:
For now I'm going to just use the full wrappers for all versions of NumPy. This is actually what the CI has been testing all along. However, we might consider using a much reduced wrapper for NumPy 2.0+, especially if it can make a difference for performance.

lucascolley (approved):
thanks Aaron, LGTM!

lucascolley (commented):
I think you need to adjust https://github.com/data-apis/array-api-compat/blob/158622e4c54eb4a8a99822cd475a3c331a254069/tests/_helpers.py#L6-L10

PR #172, opened 2024-08-01
Future-proof JAX array API import

jakevdp:
As of https://github.com/google/jax/pull/22818, `jax.numpy` supports the array API directly, and `jax.experimental.array_api` is deprecated. This change will ensure that `array_api_compat` works correctly with future JAX releases.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -377,7 +377,7 @@ def array_namespace(*xs, api_version=None, use_compat=None):
             elif use_compat is False:
                 import jax.numpy as jnp
             else:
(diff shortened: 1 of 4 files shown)


asmeurer:
Thanks. If you grep for `jax.experimental` there's a few other places that should be updated too.

jakevdp:
> Thanks. If you grep for `jax.experimental` there's a few other places that should be updated too.

Done - thanks!

rgommers:
> As of [google/jax#22818](https://github.com/google/jax/pull/22818), `jax.numpy` supports the array API directly, and `jax.experimental.array_api` is deprecated.

That's awesome 🎉

PR #174, opened 2024-08-06
Use conj_physical for torch.conj

### LINKED ISSUE #173: `conj` should delegate to `torch.conj_physical`
See https://github.com/scipy/scipy/pull/20772#issuecomment-2268760340

asmeurer:
torch.conj sets the conjugation bit, which breaks other libraries.

Fixes #173.


rgommers (approved):
Looks good modulo the failure. The more principled fix would be to call `torch.resolve_conj` only in the interop functions between libraries, however there is no way to wrap `__dlpack__` & co, so in practice this isn't doable from outside PyTorch.

The performance loss will be quite limited and specific to a few operations that PyTorch would otherwise be able to keep lazy, and isn't too much of a concern here I'd say.

Issue #156, opened 2024-06-26
Add is_numpy_namespace, etc. helper functions


asmeurer:
See https://github.com/scipy/scipy/issues/21054#issuecomment-2192403563

PR #178, opened 2024-08-15
add `is_*_namespace` helper functions

lucascolley:
closes gh-156

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -276,7 +276,7 @@ def is_numpy_namespace(xp) -> bool:
     is_pydata_sparse_namespace
     is_array_api_strict_namespace
     """
-    return xp.__name__ in {'numpy', _compat_module_name + '.numpy'}
+    return xp.__name__ in {'numpy', _compat_module_name() + '.numpy'}
 
 def is_cupy_namespace(xp) -> bool:
     """
@@ -296,7 +296,7 @@ def is_cupy_namespace(xp) -> bool:
     is_pydata_sparse_namespace
     is_array_api_strict_namespace
     """
-    return xp.__name__ in {'cupy', _compat_module_name + '.cupy'}
+    return xp.__name__ in {'cupy', _compat_module_name() + '.cupy'}
 
 def is_torch_namespace(xp) -> bool:
     """
@@ -316,7 +316,7 @@ def is_torch_namespace(xp) -> bool:
     is_pydata_sparse_namespace
     is_array_api_strict_namespace
     """
-    return xp.__name__ in {'torch', _compat_module_name + '.torch'}
+    return xp.__name__ in {'torch', _compat_module_name() + '.torch'}
     
 
 def is_ndonnx_namespace(xp):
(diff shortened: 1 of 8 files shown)


asmeurer:
Let's make these not depend on the name array_api_compat. Vendoring supports renaming the module. We should be able to just compare against `__name__`.

lucascolley:
> Let's make these not depend on the name array_api_compat. Vendoring supports renaming the module. We should be able to just compare against __name__.

How should we do that? Create an array and pass it to `is_*_array` or is there a cheaper way?

asmeurer:
I mean the global variable `__name__` will contain the name of the current module.

lucascolley:
sorry, I don't follow - what would one of the functions look like?

asmeurer:
Something like

```py
def _compat_module_name():
    assert __name__.endswith('.common._helpers')
    return __name__.removesuffix('.common._helpers')
```

lucascolley:
thanks, done

array_api_compat/common/_helpers.py
```diff
@@ -255,6 +254,166 @@ def is_array_api_obj(x):
...
+    array_namespace
+    is_cupy_namespace
+    is_torch_namespace
+    is_ndonnx_namespace
+    is_dask_namespace
+    is_jax_namespace
+    is_pydata_sparse_namespace
+    is_array_api_strict_namespace
+    """
+    return xp.__name__ in {'numpy', _compat_module_name + '.numpy'}
```
betatim, line 279: What happens when ``xp` is `numpy.array_api` from previous numpy releases? If it is wrapped I assume it will return true, but unsure for unwrapped `numpy.array_api`. (same comment applies to `cupy.array_api`)

lucascolley: Indeed I don't think this will be captured - how is scikit-learn using `numpy.array_api`? In SciPy we just use array-api-strict now.

betatim: I was thinking about users who use numpy<2 with scikit-learn. I don't know what happens then, my guess is they end up being wrapped? I thought that it would be as easy as adding `numpy.array_api` to the set to catch it, so didn't go and investigate.

So at least 50% of my question is "for my own education"

lucascolley: > I was thinking about users who use numpy<2 with scikit-learn. I don't know what happens then, my guess is they end up being wrapped?

Users should not be passing arrays from `numpy.array_api` - a numpy<2 user just needs to pass a normal NumPy array, and `array_api_compat.numpy` handles the wrapping. `numpy.array_api` was purely for testing purposes, and that is array-api-strict now.

betatim: Probably fine to skip it then? "mark as resolved"?

asmeurer: numpy.array_api would never be wrapped. It's not really appropriate to consider it a "numpy" namespace because even though it lived in numpy it had a separate array object (this confusion is one reason it was split out).

asmeurer:
Can we add some simple tests for this and also something in the vendoring test to make sure that works?

lucascolley:
Okay. Are there existing test for the `is_*_array` functions or should I add tests for those too?

asmeurer:
Yes, those are tested in test_common.py

asmeurer:
One more change and this can be merged: can you add these functions to the API docs?

lucascolley:
> can you add these functions to the API docs?

done (I think)!

PR #191, opened 2024-10-28
Update __array_api_version__ to 2023.12

asmeurer:


asmeurer:
Note that there are a few things that aren't fully covered in the test suite, namely `tile` and `searchsorted` aren't tested very well. I believe dask has some NotImplementedErrors there, but NumPy, CuPy, and Torch *should* (as far as I can see) be OK, since they just copied NumPy and the standard is based on NumPy. See https://github.com/data-apis/array-api-compat/issues/127

rgommers:
Defaulting to the most recent released version of the standard once test coverage for that version is (more or less) complete seems like a good idea to me, so +1 for this to go in when CI is happy.

If some users were relying on this in their CI without setting explicitly the version they're testing, that's easy enough to fix on their end.

PR #199, opened 2024-11-15
TST: enable the linalg extension for torch

ev-br:


asmeurer:
Oh wow good catch. I had no idea this was still being disabled (in the future, we should exclusively use the skips/xfails files to skip tests).

PR #201, opened 2024-11-19
Conjugate the first argument in vecdot

### LINKED ISSUE #200: BUG: vecdot does not conjugate first element
The implementation for `vecdot` should conjugate its first argument for complex numbers, but unfortunately the present implementation does not: https://github.com/data-apis/array-api-compat/blob/8d3f5d55cacabcbd8bc9394340d8dfdfc1fcb076/array_api_compat/common/_aliases.py#L469-L483

p.s. Found this while implementing a [new quantity class](https://github.com/astropy/quantity-2.0) that uses the array API to handle any type of array that can be passed in -- `array_api_compat` has been a great boon!

p.s.2 I found the problem for `dask`, which actually seems to be able to handle `np.vecdot` fine (and give the correct answer).

asmeurer:
Note that PyTorch uses its own vecdot directly for complex inputs,, which already does conjugation.

This is currently untested by the test suite (data-apis/array-api-tests#312)

Fixes #200.

### DIFF
--- a/array_api_compat/common/_aliases.py
+++ b/array_api_compat/common/_aliases.py
@@ -479,7 +479,7 @@ def vecdot(x1: ndarray, x2: ndarray, /, xp, *, axis: int = -1) -> ndarray:
     x2_ = xp.moveaxis(x2, axis, -1)
     x1_, x2_ = _broadcast(x1_, x2_)
 
-    res = x1_[..., None, :] @ x2_[..., None]
+    res = xp.conj(x1_[..., None, :]) @ x2_[..., None]
     return res[..., 0, 0]
 
 # isdtype is a new function in the 2022.12 array API specification.


array_api_compat/common/_aliases.py
```diff
@@ -479,7 +479,7 @@ def vecdot(x1: ndarray, x2: ndarray, /, xp, *, axis: int = -1) -> ndarray:
     x2_ = xp.moveaxis(x2, axis, -1)
     x1_, x2_ = _broadcast(x1_, x2_)
 
-    res = x1_[..., None, :] @ x2_[..., None]
+    res = xp.conj(x1_[..., None, :]) @ x2_[..., None]
```
ev-br, line 482: Wanted to write "this is obviously correct" but actually no:

```
>>> import array_api_strict as xp
>>> xp.conj(xp.ones(3))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/br/miniforge3/envs/array-api-tests/lib/python3.11/site-packages/array_api_strict/_elementwise_functions.py", line 382, in conj
    raise TypeError("Only complex floating-point dtypes are allowed in conj")
TypeError: Only complex floating-point dtypes are allowed in conj

```

ev-br (changes requested):
The spec stance on  rejecting real arguments to complex functions  is a bit bizzare. I realize it's probably because `xp.imag(real_array)` would otherwise have to allocate an array of zeros but still.

asmeurer:
And also you don't really ever have to worry about array-api-strict in array-api-compat. The only place it's ever used is it can be returned from `array_namespace`.

asmeurer:
I'm going to merge this now, but I'm hopeful we can get testing for this in array-api-tests before we do a release.

Issue #146, opened 2024-06-11
Helper functionality to work around how different libraries handle copies vs. mutation and/or views?


mdhaber:
Follow-up to https://github.com/data-apis/array-api-compat/issues/144#issuecomment-2161491665

Sometimes to make existing code compatible with backends that are not fully standard compliant, we would need to create copies where the original code would not.

For example, as a workaround for gh-144 (PyTorch doesn't support negative `step`), we _could_ (sometimes) make the replacement:
```python3
from array_api_compat import torch
x = torch.arange(10)
# x[::-1]  
xp.flip(x)  # ValueError: step must be greater than zero
``` 

As a workaround for JAX not supporting mutation, we *could* sometimes make replacements like:
```python3
from array_api_compat import jax
# x is an array, i is a mask with the same shape
# x[i] = 0
x = jax.where(i, 0, x)
```

However, making substitutions like this could decrease performance for array types that do support the desired operation (returning a view or mutating the original, in these cases).

Functions that perform the desired operation when possible and the substitute otherwise (e.g. https://github.com/scipy/scipy/pull/20085#issuecomment-2119307509) have been proposed. Do such things belong in `array_api_compat`?

asmeurer:
Regarding `flip(x)` vs. `x[::-1]`, note that `np.flip` does return a view. Torch only doesn't because it doesn't support views with negative strides at all. So just using `flip` is portable in the manner you suggest. It just isn't as readable (though even that could be argued), but there's no real way around that since we don't want to wrap array/tensor objects in this library. More arbitrary slices would also be more cumbersome to write, and we can definitely build a helper around that if it is needed.

What would an API for "mutate if you can" look like? 

You'd have to be really careful using any API like that. A mutation and a copy are very different things and you'd need to make sure you write code that works correctly in both instances. I guess the way to handle it is to never actually rely on mutation semantically. Rather, mutation should just be treated as an implementation detail for performance. In other words, write code that treats everything as immutable but using functions that can mutate or "copy on write" when possible.

JAX effectively does this internally, where non-aliased arrays are free to be mutated. I wonder if this request effectively amounts to "rewrite JAX on top of any array API library". If so, it might be very difficult or even impossible to achieve. For instance, even in NumPy we don't currently have the ability to fully track aliasing (views are only tracked in one direction). 

CC @rgommers

mdhaber:
> Regarding flip(x) vs. x[::-1], note that np.flip does return a view. 

[...]

PR #205, opened 2024-11-25
ENH: Test for read-only arrays

crusaderky:
New public function `is_writeable_array`, which introduce transparent support for read-only backends such as JAX (but more may be added in the future).

[EDIT] this PR also implemented an `at` function, mocking the syntax of JAX's omonymous method.
This function will be proposed in array-api-extra instead.

### DIFF
--- a/array_api_compat/__init__.py
+++ b/array_api_compat/__init__.py
@@ -1,9 +1,9 @@
 """
 NumPy Array API compatibility library
 
-This is a small wrapper around NumPy and CuPy that is compatible with the
-Array API standard https://data-apis.org/array-api/latest/. See also NEP 47
-https://numpy.org/neps/nep-0047-array-api-standard.html.
+This is a small wrapper around NumPy, CuPy, JAX, sparse and others that are
+compatible with the Array API standard https://data-apis.org/array-api/latest/.
+See also NEP 47 https://numpy.org/neps/nep-0047-array-api-standard.html.
 
 Unlike array_api_strict, this is not a strict minimal implementation of the
 Array API, but rather just an extension of the main NumPy namespace with
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -787,6 +787,7 @@ def to_device(x: Array, device: Device, /, *, stream: Optional[Union[int, Any]] 
         return x
     return x.to_device(device, stream=stream)
 
+
 def size(x):
     """
     Return the total number of elements of x.
@@ -800,6 +801,23 @@ def size(x):
     if None in x.shape:
         return None
     return math.prod(x.shape)
+
+
+def is_writeable_array(x) -> bool:
+    """
+    Return False if ``x.__setitem__`` is expected to raise; True otherwise.
+
+    Warning
+    -------
+    As there is no standard way to check if an array is writeable without actually
+    writing to it, this function blindly returns True for all unknown array types.
+    """
+    if is_numpy_array(x):
+        return x.flags.writeable
+    if is_jax_array(x) or is_pydata_sparse_array(x):
+        return False
+    return True
+
 
 __all__ = [
     "array_namespace",
@@ -821,6 +839,7 @@ __all__ = [
     "is_ndonnx_namespace",
     "is_pydata_sparse_array",
     "is_pydata_sparse_namespace",
+    "is_writeable_array",
     "size",
     "to_device",
 ]
--- a/docs/helper-functions.rst
+++ b/docs/helper-functions.rst
@@ -51,6 +51,7 @@ yet.
 .. autofunction:: is_jax_array
 .. autofunction:: is_pydata_sparse_array
 .. autofunction:: is_ndonnx_array
+.. autofunction:: is_writeable_array
 .. autofunction:: is_numpy_namespace
 .. autofunction:: is_cupy_namespace
 .. autofunction:: is_torch_namespace
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -5,7 +5,7 @@ from array_api_compat import (  # noqa: F401
     is_dask_namespace, is_jax_namespace, is_pydata_sparse_namespace,
 )
 
-from array_api_compat import is_array_api_obj, device, to_device
+from array_api_compat import device, is_array_api_obj, is_writeable_array, to_device
 
 from ._helpers import import_, wrapped_libraries, all_libraries
 
@@ -72,6 +72,24 @@ def test_xp_is_array_generics(library):
         if is_func(x0):
             matches.append(library2)
     assert matches in ([library], ["numpy"])
+
+
+@pytest.mark.parametrize("library", all_libraries)
+def test_is_writeable_array(library):
+    lib = import_(library)
+    x = lib.asarray([1, 2, 3])
+    if is_writeable_array(x):
+        x[1] = 4
+    else:
+        with pytest.raises((TypeError, ValueError)):
+            x[1] = 4
+
+
+def test_is_writeable_array_numpy():
+    x = np.asarray([1, 2, 3])
+    assert is_writeable_array(x)
+    x.flags.writeable = False
+    assert not is_writeable_array(x)
 
 
 @pytest.mark.parametrize("library", all_libraries)


asmeurer:
This seems related to the suggestions at https://github.com/data-apis/array-api-compat/issues/146. Ping @mdhaber @lucascolley to check if these helpers will be useful in SciPy.

lucascolley:
x-ref https://github.com/data-apis/array-api/issues/609, cc @jakevdp @rgommers

mdhaber:
Yes, it looks like it!

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +801,174 @@ def size(x):
...
+
+    See Also
+    --------
+    https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html
+    """
+    if is_jax_array(x):
+        return x.at
+    if is_numpy_array(x) and not x.flags.writeable:
+        x = x.copy()
+    return _InPlaceAt(x, idx)
```
lucascolley, line 870: So if I understand correctly, this implementation works for:
- JAX
- NumPy
- Mutable arrays (arrays for which the in-place updates work)

It doesn't work for immutable arrays coming from other libraries, for the reasons discussed in https://github.com/data-apis/array-api/issues/845.

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +801,174 @@ def size(x):
...
+    whereas y == z == [2, 3, 0] when x is writeable!
+
+    See Also
+    --------
+    https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html
+    """
+    if is_jax_array(x):
+        return x.at
+    if is_numpy_array(x) and not x.flags.writeable:
+        x = x.copy()
```
lucascolley, line 869: perhaps we would want an optional kwarg to error instead of copy here? I suppose the semantics for `copy=True/None/False` used elsewhere in the standard could fit?

lucascolley (commented):
yep, looks useful! The diffs will be dauntingly large I suppose, but hopefully trivial enough!

array_api_compat/common/_helpers.py
crusaderky: It works for all arrays explicitly listed in array-api-compat, including numpy subclasses.
It won't work for immutable arrays from other, unknown libraries - in fact, they won't be even recognized as immutable. Without a standardized way for `is_immutable_array` to detect such a use case (`__array_writeable__` interface?) I would not know how to implement it.

array_api_compat/common/_helpers.py
crusaderky: This is a great idea

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +802,257 @@ def size(x):
...
+    Examples
+    --------
+    Given either of these equivalent expressions::
+
+      x = at(x)[1].add(2, copy=None)
+      x = at(x, 1).add(2, copy=None)
+
+    If x is a JAX array, they are the same as::
+
+      x = x.at[1].add(x)
```
jakevdp, line 859: Typo:
```suggestion
      x = x.at[1].add(2)
```

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +802,257 @@ def size(x):
...
+        self.x = x
+        self.idx = idx
+
+    def __getitem__(self, idx):
+        """
+        Allow for the alternate syntax ``at(x)[start:stop:step]``,
+        which looks prettier than ``at(x, slice(start, stop, step))``
+        and feels more intuitive coming from the JAX documentation.
+        """
+        self.idx = idx
```
jakevdp, line 903: I think this should error if `self.idx` is already set, otherwise people may expect `at(x)[0][1].set(2)` to be the same as `x[0][1] = 2`

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +802,257 @@ def size(x):
...
+            x[idx] = x[idx] + y
+        Note that this is not the same as 
+            operator.iadd(x[idx], y)
+        Consider for example when x is a numpy array and idx is a fancy index, which
+        triggers a deep copy on __getitem__.
+        """
+        res, x = self._common(at_op, y, **kwargs)
+        if res is not None:
+            return res
+        x[self.idx] = elwise_op(x[self.idx], y)
```
jakevdp, line 998: One issue here: there are different semantics between JAX and NumPy for repeated indices as this is currently implemented. For example:
```python
import numpy as np
x = np.ones(2)
i = np.array([0, 0])
x[i] *= 2
x  # array([2., 1.])
```
```python
import jax.numpy as jnp
x = jnp.ones(2)
x.at[i].mul(2)  # Array([4., 1.], dtype=float32)
```
We could fix this by dispatching the NumPy version to `np.multiply.at`, or maybe we could document that repeated indices lead to different behavior on different backends.

crusaderky:
> The diffs will be dauntingly large I suppose

Do you mean the diffs inside scipy? Yes they will. Not much that can be done about it I'm afraid.

I gave this some more thought and reworked the PR.
Now

- `where` is a wrapper around `xp.where`, which adds copy=True/False/None (defaults to True).
- `at` is a simple wrapper around `jnp.ndarray.at`, which adds the copy=True/False/None (defaults to True) parameter. When x is not a JAX array, `at` behaves the same as JAX by default (e.g. deep-copy everything).

In scipy and similar libraries, we should replace all instances of
```python
x[idx] += y
```
with
```python
x = at(x, idx).add(y, copy=None)
```
(read below for the masked use case).

Offline comment by @rgommers :
> this targets the wrong repo. `array-api-compat` is a compatibilty layer for things that are actually in the standard, or about to land there. `array-api-extra` is for things on top.

There is discussion elsewhere that something should be pushed into the standard.
I would like to suggest exactly that for this PR. I think we should add to the API standard:
- An `at[]` method, as already implemented in JAX plus the copy=True optional parameter, plus additional backend-specific kwargs (e.g. the standard would omit things like `indices_are_sorted`). The behaviour for out-of-bounds indices should likewise be left to undefined.
-  adding the copy=True parameter to `where`. 

## Masked JAX arrays
I've deliberately omitted special-casing for masked JAX arrays, unlike in @rgommers 's POC here https://github.com/scipy/scipy/compare/main...rgommers:scipy:array-types-inplace-ops?expand=1.

IMHO, the problem with the POC code e.g.
```python
    if is_jax(xp):
        if hasattr(idx, 'dtype') and xp.isdtype(idx.dtype, 'bool'):
            x = xp.where(idx, x + y, x)
        else:
            x = x.at[idx].add(y)
    else:
        x[idx] += val
```
is that it produces incorrect results when y has shape other than `()` or `(1, )`.

This strongly feels to me less of a generic problem with read-only arrays (which array-api-compat and scipy should cater for) and more of a JAX-specific quirk (needing to know all shapes in advance).
So I think we should implement the above snippet as a rewrite rule inside `jax.jit`:
- if idx is a JAX array, and
- y has shape `()` or `(1, )`,
- then rewrite `x.at[idx].add(y)` as `jnp.where(idx, x + y, x)`

Until that happens, scipy code will work with JAX, but it will break when jitted.

CC @jakevdp

lucascolley:
> this targets the wrong repo. array-api-compat is a compatibilty layer for things that are actually in the standard, or about to land there. array-api-extra is for things on top.

I think I disagree. Everything in array-api-extra works with any standard-compatible library, whereas this implementation will not work with immutable arrays from libraries not covered by array-api-compat (there doesn't seem to be any way it could).

Meanwhile, none of the helper functions in array-api-compat (https://data-apis.org/array-api-compat/helper-functions.html) are in the standard - that's where this PR is targeting.

I suppose the worry is that this is effectively arguing for the standardisation of `at` (as @crusaderky has made explicit) by demonstrating that it can be implemented for all libraries currently supported by array-api-compat. I don't think it would be the end of the world for this to stay in array-api-compat though, rather than being pushed into the array libraries themselves, as long as there isn't a library for which support is too complicated.

array_api_compat/common/_helpers.py
crusaderky: Wouldn't that be a numpy-specific hack? In order to make it generic for other writeable, non-numpy arrays, we would need to amend the `__array_ufunc__` API to add `method="at"`?

I'd definitely prefer documenting that such a behaviour is undefined. 
I can't honestly think of any sane use case for repeated indices in updates?

jakevdp: It’s useful for things like implementing `bincount` or `histogram` in a vectorized manner

asmeurer:
I think it's a fuzzy zone whether this belongs in compat or in extra. The scope for array-api-compat is spelled out here https://data-apis.org/array-api-compat/#scope. On the one hand, the `at` helper is much like the `device` or `size` helpers already in compat. It converts what would normally be a method of an array object to a helper function, since methods cannot be wrapped. But you could also argue it's not quite the same as that since the "method" being wrapped here isn't really `__setitem__` but rather `.at`, which isn't in the standard (yet). So one could also argue it's out of scope for array-api-compat. I think the biggest concern here is whether adding this here would be jumping the gun a bit for something that might later be added to the standard itself. 

`iwhere` feels more like something that you could make this argument about, and I do wonder if it ought to be here.

`is_writeable_array` seems fine, although the main concern there as noted is that it's hard to make it generic against libraries that aren't covered by array-api-compat.

(by the way, we should add a reference to array-api-extra to that scope section)

array_api_compat/common/_helpers.py
asmeurer: Would attempting to mutate the array as a fallback work, or could that lead to undesired side-effects?

```py
# Something along these lines, but more robust
a = x[0]
try:
    x[0] = 0
    x[0] = a
    mutable = True
except:
    mutable = False
```

lucascolley:
> But you could also argue it's not quite the same as that since the "method" being wrapped here isn't really __setitem__ but rather .at, which isn't in the standard (yet). So one could also argue it's out of scope for array-api-compat. I think the biggest concern here is whether adding this here would be jumping the gun a bit for something that might later be added to the standard itself.

IMO the bar is finding agreement on a sensible API in https://github.com/data-apis/array-api/issues/609 (and here). But we don't need to have any confidence that existing array libraries will adopt it, since the implementations can live in array-api-compat if need be.

crusaderky:
> iwhere feels more like something that you could make this argument about, and I do wonder if it ought to be here.

Note that I renamed it to `where` now. It's identical to the numpy one with the added parameter copy=True|False|None.
I'm not sure I will need it in scipy either - I suspect it would be redundant with the `@jax.jit` change I suggested above (https://github.com/data-apis/array-api-compat/pull/205#issuecomment-2501023419) - so I'm inclined to take it out of this PR and potentially reopen the discussion in a later PR if it turns out I do need it.

array_api_compat/common/_helpers.py
crusaderky: Caveat added

array_api_compat/common/_helpers.py
crusaderky: I think that reimplementing `at[]` without the ability to update and without specific knowledge of the library at hand would be extraordinarily inefficient.

Consider `at(x, 0).add(1)`. Implementing it on top of `where` would be extremely expensive; an alternative would be to break the array down and rebuild it with `concatenate` and `stack`, which would be very complicated and probably fragile.

The snippet of code you just wrote would perform very poorly in many cases. Consider:
- a library like dask: the `x[0] = a` line adds extra labour into the graph, which is very likely nontrivial
- any library where the memory transparently moves from device to host and vice versa would likely cause either a page or the whole array to move from device to host (and vice versa).

A somewhat slightly more reasonable alternative would be to

1. blindly try the update
2. if the update fails, try if by any chance the library has a `at[]` method with exactly the same API as JAX

At the moment we're only doing (1).
My personal opinion would be to explicitly cater for these libraries if and when they crop up.
Is this whole discussion hypothetical, or do we know of a specific read-only library other than JAX?

crusaderky: I just realised that sparse.COO does not support updates. Has there been any discussion about it?

crusaderky:
I think this is ready for a second round of consultation.
As mentioned above, I've excised where/iwhere from this PR.

## sparse
`sparse` is read-only and doesn't support `.at` or similar tools to my understanding, so none of it works at the moment. My opinion is that I should XFAIL everything now and then discuss implementing an `.at` method in the `sparse` library itself. 

CC @hameerabbasi 

## JAX apply() and ufuncs
I can't seem to make JAX's apply() work:
```python
>>> import jax.numpy as jnp
>>> import numpy as np
>>> a = jnp.array([1,2,3])
>>> a.at[:2].apply(np.negative)
jax.errors.TracerArrayConversionError: The numpy.ndarray conversion method __array__() was called on traced array with shape int32[]
```
This works without JIT...
```python
>>> a.at[:2].set(np.negative(a[:2]))
Array([-1, -2,  3], dtype=int32)
```
... but it crashes with `@jax.jit` with the same error as above.
I think the giveaway is that intermediate output of the ufunc applied to the JAX array is a plain numpy array:
```python
>>> np.negative(a)
array([-1, -2, -3], dtype=int32)
```

Given
- the above problems, and
- the fact that there is a straightforward workaround with set() that works as long as plain ufunc application is fixed in JAX, and
- that we need to implement the workaround _anyway_ for cupy, torch, and dask, as none of them implement ufunc.at, and
- that the key feature of apply, that is the accumulation behaviour on repeated indices, won't be consistent anyway across backends, and
- its limitation to unary ufuncs (unlike ufunc.at, which supports a second parameter),

**I'm inclined to completely remove apply() from the API.**

## PyTorch ufuncs
Applying a ufunc to a torch array raises a warning, which makes the tests for min, max, and apply fail.
Should I suppress the warning in the test suite?
```python
```python
>>> import torch
>>> import numpy
>>> a = torch.asarray([1,2,3])
>>> numpy.negative(a)
DeprecationWarning: __array_wrap__ must accept context and return_scalar arguments (positionally) in the future. (Deprecated NumPy 2.0)
tensor([-1, -2, -3])
```

array_api_compat/common/_helpers.py
```diff
@@ -91,7 +92,7 @@ def is_cupy_array(x):
     import cupy as cp
 
     # TODO: Should we reject ndarray subclasses?
-    return isinstance(x, (cp.ndarray, cp.generic))
+    return isinstance(x, cp.ndarray)
```
crusaderky, line 95: `cp.generic is np.generic`. So pure-numpy use cases would take a cupy-specific code path.
Need to move this to a separate PR.

array_api_compat/common/_helpers.py
```diff
@@ -801,6 +803,262 @@ def size(x):
...
+        """ufunc.at(x, idx)"""
+        if is_cupy_array(self.x) or is_torch_array(self.x) or is_dask_array(self.x):
+            # ufunc.at not implemented
+            return self.set(ufunc(self.x[self.idx]), **kwargs)
+
+        res, x = self._common("apply", ufunc, **kwargs)
+        if res is not None:
+            return res
+        ufunc.at(x, self.idx)
+        return x
```
crusaderky, line 1009: IMHO we should delete this whole method (full rationale: https://github.com/data-apis/array-api-compat/pull/205#issuecomment-2504865475)

tests/test_at.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def assert_array_equal(a, b):
+    if is_pydata_sparse_array(a):
+        a = a.todense()
+    elif is_cupy_array(a):
+        a = a.get()
+    elif is_dask_array(a):
+        a = a.compute()
+    np.testing.assert_array_equal(a, b)
```
crusaderky, line 29: This feels like a problem someone else already dealt with elsewhere.
Did I reinvent the wheel?

tests/test_at.py
```diff
@@ -0,0 +1,166 @@
...
+    if is_jax_array(x) and isinstance(idx, (list, tuple)):
+        pytest.skip("JAX fancy indices must always be arrays")
+    if is_pydata_sparse_array(x) and is_pydata_sparse_array(idx):
+        pytest.skip("sparse fancy indices can't be sparse themselves")
+    if is_torch_array(x) and isinstance(idx, np.ndarray) and idx.dtype.kind == "u":
+        pytest.skip("torch does not support unsigned integer fancy indices")
+    if is_dask_array(x) and isinstance(idx, tuple):
+        pytest.skip("dask does not support tuples; only lists or arrays")
+    if isinstance(idx, tuple) and not tuple_index:
+        pytest.skip("tuple indices must always be wrapped in a tuple")
```
crusaderky, line 133: This forest of special cases is awful. 
Suggestions to clean it away?

asmeurer:
Looking a little closer at this, my personal opinion is that it would make more sense at the moment to put `at` in the array-api-extra library rather than here in array-api-compat. There is a lot more API surface to it than I had initially realized, i.e., all the methods on the object. array-api-extra is supposed to be a little looser on potential future API breakages (although we haven't completely figured out how that will work; I proposed making it vendor-only, but there are other ways to do this too).

It also looks like this is trying to use NumPy functions uniformly, regardless of whether the underlying library is NumPy. That's going to work for specific libraries that implement the NumPy interop (`__array_function__` and so on), but it goes against the spirit of the array API. In particular, it limits how much this can be extended to other libraries beyond the "known" ones supported here. And this would never be part of the API if it were at some point standardized. 

`is_writable_array` could make sense as a helper here. It would be nice if it at least optionally attempted to be generic for unknown libraries, but also it makes sense to not try to do that if it's too hard or has too many pitfalls. This definitely seems like an API that should be added to the array API (probably in the inspection namespace). I would at least propose it over on the array-api repo.

tests/test_at.py
asmeurer: I don't know, but each of these seems like something that could come up as a problem in practice. Might be worth documenting.

asmeurer:
Just to be clear, someone else (probably @rgommers) should make the actual decision on whether or not this belongs here or elsewhere.

crusaderky:
> It also looks like this is trying to use NumPy functions uniformly, regardless of whether the underlying library is NumPy.

I've replaced np.minimum with xp.minimum - is this what you were referring to?

asmeurer:
Yes, that was what I meant. And also the apply function which you already mentioned you want to remove.

jakevdp:
> I can't seem to make JAX's apply() work:

In JAX, you can apply JAX functions, not NumPy functions:
```python
>>> import jax.numpy as jnp
>>> a = jnp.array([1,2,3])
>>> a.at[:2].apply(jnp.negative)
Array([-1, -2,  3], dtype=int32)
```

That said, I don't think `apply` is all that important a function for this helper, so I wouldn't be opposed to leaving it out.

lucascolley:
agreed with leaving out `apply` for this PR. @crusaderky you may be interested in `xarray`'s work on generic ufuncs which apply functions from the respective underlying library: https://github.com/pydata/xarray/pull/9776.

hameerabbasi:
> `sparse` is read-only and doesn't support `.at` or similar tools to my understanding, so none of it works at the moment. My opinion is that I should XFAIL everything now and then discuss implementing an `.at` method in the `sparse` library itself.

That seems like a good plan, yes.

rgommers:
> Applying a ufunc to a torch array raises a warning, which makes the tests for min, max, and apply fail. Should I suppress the warning in the test suite?

Mixing `numpy` and `pytorch` (or any two libraries) like that is some you shouldn't have to worry about, this is fundamentally not supportable. And that is also documented, from https://data-apis.org/array-api/latest/assumptions.html: _"Importantly, array libraries are not assumed to be aware of each other, or of a common array-specific layer."_

If the test suite has such mixing in it at the moment, it's probably by accident I'd think.

array_api_compat/common/_helpers.py
crusaderky: Blocked by https://github.com/data-apis/array-api-compat/pull/207

rgommers:
Re scope, I agree with @asmeurer's last comment above: the `is_writeable` utility seems pretty clearly in scope, and it seems pretty clear cut to me that `at` is not in scope. That is a lot of API surface and not a utility function, hence it cannot go in `-compat` unless it will be standardized. And standardizing it is a big lift, and not the most desirable outcome in principle. We can discuss that of course (https://github.com/data-apis/array-api/issues/609 is better than this PR), but the better outcome is making item/slice assignment robust against views in NumPy et al, so that it becomes possible for JAX to match NumPy. The engineering work here is in NumPy/CuPy/PyTorch primarily.

I'll also note that the scope for `-compat` has been very well documented (and policed) by @asmeurer for several years.  
For `-extra`, which is brand new, and we've _started_ with pure Python 100% portable functions, but _also discussed_ that the type of functionality that needs per-library tweaks may live there - we punted on that to get the ball rolling quickly. @lucascolley I cannot quite locate that discussion from a couple of months ago, but I suspect you can?

My ideal outcome here is probably something like:

1. merge `is_writable_array` here
2. merge `at` in `array-api-extra`
3. Pull both updated packages into SciPy
4. Start using `at` to make a lot of JAX functionality work
5. (assuming we're happy) start doing the same in scikit-learn et al.
6. Enjoy the benefits of using JAX after the next releases of the projects that do (4), (5) :)

... (longer time scale) ...

7. Get to actual portable behavior without the `at` helper. Ideally by making https://data-apis.org/array-api/latest/design_topics/copies_views_and_mutation.html more strict (mutating aliased memory becomes not allowed) and all libraries adhering to that. Alternatively, explore whether `at` can be standardized (a much harder sell, because it effectively deprecates item/slice assignment permanently - this was a deal breaker for adoption before, and there's probably >100x more code out there using item/slice assignment then there is using `at`).

crusaderky:
@rgommers moving this would add a hard dependency in `array-api-extra` on `array-api-compat`, as at the very least `at` needs to call `is_writeable_array`; is that ok?

rgommers:
> @rgommers moving this would add a hard dependency in `array-api-extra` on `array-api-compat`, as at the very least `at` needs to call `is_writeable_array`; is that ok?

Ah 🤔  - I think that that will become unavoidable soon enough, but it does complicate vendoring if the versions of two packages start to become linked. I suggest to wait for a bit to give @lucascolley a change to reply to both my message above and this question, since he is the main maintainers of `-extra`.

lucascolley:
your plan sounds good @rgommers , if not for the complicated vendoring situation. Without that complication, I think it would be okay to partition the array-api-extra API into one part which works with any compatible library, and another part which only works for libraries supported by array-api-compat.

I discuss the complicated vendoring situation in https://github.com/data-apis/array-api-extra/issues/6. The PR cross-linked to that issue contained the initial discussion on whether array-api-compat should be a dependency or not.

lucascolley:
Overall I feel like it is less messy to partition the array-api-compat API into standard and non-standard things, but just be very picky about what can make it into the non-standard part. Picky in the sense of https://github.com/data-apis/array-api-compat/pull/205#issuecomment-2501839881.

crusaderky:
> Overall I feel like it is less messy to partition the array-api-compat API into standard and non-standard things

How would that look like? Separate section in the sphinx docs and separate module for the implementation?

crusaderky:
> partition the array-api-extra API into one part which works with any compatible library, and another part which only works for libraries supported by array-api-compat.

How would the import look like in `array-api-extra`? 
```python
from array_api_compat import writeable_array
```
doesn't look right for the vendored use case? Or would array-api-extra itself vendor array-api-compat internally? (and in turn scipy vendors array-api-extra, so scipy would end up having two vendored copies of array-api-compat)

rgommers:
> doesn't look right for the vendored use case?

Agreed. https://github.com/data-apis/array-api-extra/issues/6#issuecomment-2506673209 has a suggestion that should work I think.

crusaderky:
In https://github.com/data-apis/array-api-extra/issues/6#issuecomment-2507674379 I broke down how adding a dependency to array-api-compat in array-api-extra would look like; it looks quite messy to me...

crusaderky:
Removed `at`. Will open a PR in array-api-extra.

This PR is ready to be reviewed and merged.
- CI failures are unrelated and should be fixed by https://github.com/data-apis/array-api-tests/pull/325
- `at` will also depend on https://github.com/data-apis/array-api-compat/pull/207

array_api_compat/__init__.py
```diff
@@ -1,9 +1,9 @@
 """
 NumPy Array API compatibility library
 
-This is a small wrapper around NumPy and CuPy that is compatible with the
-Array API standard https://data-apis.org/array-api/latest/. See also NEP 47
-https://numpy.org/neps/nep-0047-array-api-standard.html.
+This is a small wrapper around NumPy, CuPy, JAX, sparse and others that is
+compatible with the Array API standard https://data-apis.org/array-api/latest/.
```
lucascolley, line 5: nit

```suggestion
This is a small wrapper around NumPy, CuPy, JAX, sparse and others that are
compatible with the Array API standard https://data-apis.org/array-api/latest/.
```

array_api_compat/__init__.py
```diff
@@ -1,9 +1,9 @@
...
+See also NEP 47 https://numpy.org/neps/nep-0047-array-api-standard.html.
```
lucascolley, line 6: (unrelated to this PR, but probably no need to link to that NEP now)

lucascolley (approved):

array_api_compat/__init__.py
```diff
@@ -17,6 +17,6 @@
 this implementation for the default when working with NumPy arrays.
 
 """
-__version__ = '1.9.1'
+__version__ = '1.10.0dev0'
```
crusaderky, line 20: Needed to work around https://github.com/prefix-dev/pixi/issues/2677
Without this we can't build array-api-extra until the next release: https://github.com/data-apis/array-api-extra/pull/53#discussion_r1876022501

array_api_compat/__init__.py
```diff
@@ -17,6 +17,6 @@
 this implementation for the default when working with NumPy arrays.
 
 """
-__version__ = '1.9.1'
+__version__ = '1.10.0.dev0'
```
crusaderky, line 20: Needed to work around https://github.com/prefix-dev/pixi/issues/2677
Without this we can't build array-api-extra until the next release: https://github.com/data-apis/array-api-extra/pull/53#discussion_r1876022501

crusaderky:
@lucascolley this is ready to be merged

lucascolley:
Sounds good, but I don't have merge permissions on this repo.

array_api_compat/__init__.py
ev-br: Can you submit this separately please? If this is needed for downstream, it can be merged quickly, I suppose.

array_api_compat/common/_helpers.py
ev-br: gh-207 is merged, so can rebase on main

ev-br:
Would be helpful to summarize the current scope. So ATM, this PR only adds a single public function, in the vein of https://github.com/data-apis/array-api/issues/845, correct?
Then it'd be helpful to spell the plan w.r.t. the spec: if the spec gets something in `capabilities`, will it obviate the need for this function, or will this function wrap it or....?

array_api_compat/common/_helpers.py
crusaderky: already done

array_api_compat/__init__.py
crusaderky: https://github.com/data-apis/array-api-compat/pull/216

crusaderky:
Ready to merge

ev-br:
Thanks @crusaderky, all.

PR #207, opened 2024-11-28
BUG: cp.generic is an alias for np.generic

crusaderky:
Fix bug where plain numpy objects e.g. np.int64(0) would take cupy-specific code paths.

This is a blocker for implementing `at` function in array-api-extra.

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -91,7 +91,7 @@ def is_cupy_array(x):
     import cupy as cp
 
     # TODO: Should we reject ndarray subclasses?
-    return isinstance(x, (cp.ndarray, cp.generic))
+    return isinstance(x, cp.ndarray)
 
 def is_torch_array(x):
     """
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -55,6 +55,25 @@ def test_is_xp_namespace(library, func):
     assert is_func(lib) == (func == is_namespace_functions[library])
 
 
+@pytest.mark.parametrize('library', all_libraries)
+def test_xp_is_array_generics(library):
+    """
+    Test that scalar selection on a xp.ndarray always returns
+    an object that matches with exactly one among the is_*_array
+    function of the same library and is_numpy_array.
+    """
+    lib = import_(library)
+    x = lib.asarray([1, 2, 3])
+    x0 = x[0]
+
+    matches = []
+    for library2, func in is_array_functions.items():
+        is_func = globals()[func]
+        if is_func(x0):
+            matches.append(library2)
+    assert matches in ([library], ["numpy"])
+
+
 @pytest.mark.parametrize("library", all_libraries)
 def test_device(library):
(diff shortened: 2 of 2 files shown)


crusaderky:
Test failures are unrelated

ev-br:
What would be a scenario where cupy aliasing array scalars matters?  I just tried running https://github.com/data-apis/array-api-compat/blob/main/test_cupy.sh locally, and it passes with or without.

crusaderky:
> What would be a scenario where cupy aliasing array scalars matters? 

Not that I'm aware of.
If there are any cases where the output of `is_cupy_array(cp.somefunction(...))` needs to return True, even if the output is a `np.generic`, then those cases will need to be looked into.

To me this looks like a copy-paste error from `is_numpy_array`.

ev-br:
So if there's no use case, what are the reasons for this then?

ev-br:
> But it is a big problem if is_cupy_array(numpy.asarray([1])[0]) returns True - which is what happens now in main.

That's bad indeed if that happens, but... does it?

```
In [5]: is_cupy_array(np.asarray([1])[0])
Out[5]: False
```

This is on main, after

```
$ pip install -e .
Obtaining file:///home/br/repos/array-api-compat
  Preparing metadata (setup.py) ... done
Installing collected packages: array_api_compat
  DEPRECATION: Legacy editable install of array_api_compat==1.9.1 from file:///home/br/repos/array-api-compat (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
  Running setup.py develop for array_api_compat
Successfully installed array_api_compat-1.9.1
(array-api-tests) br@gonzales:~/repos/array-api-compat$ git log
commit edde2e04a984b65dc299321ea5c0cda33f1f9e5d (HEAD -> main, upstream/main, upstream/HEAD)
Merge: d9df003 484f725
Author: Evgeni Burovski <evgeny.burovskiy@gmail.com>
Date:   Thu Dec 12 18:58:34 2024 +0200

    Merge pull request #213 from ev-br/torch_allclose
    
    MAINT: avoid a DeprecationWarning from torch

```

What am I missing?

crusaderky:
> What am I missing?

You're missing cupy.

```python
In [1]: from array_api_compat import is_cupy_array
In [2]: import numpy as np
In [3]: g = np.asarray([0])[0]
In [4]: is_cupy_array(g)
Out[4]: False
In [5]: import cupy
In [6]: is_cupy_array(g)
Out[6]: True
```

ev-br (approved):
Ouf, yes indeed, `is_cupy_array(....)` is False unless `cupy` has been imported. So yes, a clear bug with a clear fix. Thank you @crusaderky

Issue #209, opened 2024-12-05
`test_asarray_copy[dask.array]` fails on CI with dask==2024.12


ev-br:
Observed in https://github.com/data-apis/array-api-compat/pull/208 :

```
FAILED tests/test_common.py::test_asarray_copy[dask.array] - assert False
 +  where False = <function test_asarray_copy.<locals>.<lambda> at 0x7ff78986c220>(dask.array<getitem, shape=(), dtype=float32, chunksize=(), chunktype=numpy.ndarray> == 0.0)
```

Note that it 
- fails on python 3.11 job (https://github.com/data-apis/array-api-compat/actions/runs/12174769944/job/33957309550?pr=208) which installs dask 2024.12 (released 4 Dec 2024), and
- passes on python 3.9 which installs dask 2014.08 (https://github.com/data-apis/array-api-compat/actions/runs/12174769944/job/33957305292?pr=208)

ev-br:
The failing test is 


```
       # Use the standard library array to test the buffer protocol

       # ... snip ...
    
        a = array.array('f', [1.0])
        b = asarray(a, copy=None)
        assert is_lib_func(b)
        a[0] = 0.0
        if library == 'cupy':
            # A copy is required for libraries where the default device is not CPU
            assert all(b[0] == 1.0)
        else:
           assert all(b[0] == 0.0)      # <<<<< here
```

A prime suspect is https://github.com/dask/dask/pull/11524.

ev-br:
This is indeed a breaking change in dask: 2024.12: `dask.from_asarray` does

3588 ->	    if is_arraylike(x) and hasattr(x, "copy"):
3589 	        x = x.copy()

Here `x` is a numpy array which we constructed from an `array.array` instance at https://github.com/data-apis/array-api-compat/blame/main/array_api_compat/dask/array/_aliases.py#L142

So the meaning of `array_api_compat.dask.array.asarray(..., copy=None)` changes: there is no way it could reuse the existing buffer anymore, and copy=None is equivalent to copy=True. 

Does this sound right @asmeurer ?

ev-br:
Also ping @lithomas1 : would be able to weigh in on the dask.array changes?

lithomas1:
Yeah, part of the reason we wrap ``asarray`` is because dask's asarray doesn't have a copy keyword.
(I do a manual copy in there for copy=True)

I think your analysis is right that dask is always doing a copy.

dcherian:
[...]

Issue #210, opened 2024-12-05
dask==2024.12 tests hang on CI


ev-br:
https://github.com/data-apis/array-api-compat/actions/runs/12174770255/job/33957310078?pr=208 runs for 3 hrs already, stuck at 

```
array_api_tests/test_creation_functions.py::test_empty_like PASSED       [  3%]
```

ev-br:
gh-211 added a skip for an offending test, so no need to track this separately.

PR #211, opened 2024-12-05
BUG: dask.array: asarray(..., copy=None) copies in dask==2024.12

ev-br:
closes gh-209

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -206,7 +206,8 @@ def test_asarray_copy(library):
     a[0] = 0.0
     if library in ('cupy', 'dask.array'):
         # A copy is required for libraries where the default device is not CPU
-        # dask made a breaking change in 2024.12: copy=None copies
+        # dask changed behaviour of copy=None in 2024.12 to copy;
+        # this wrapper ensures the same behaviour in older versions too.
         # https://github.com/dask/dask/pull/11524/
         assert all(b[0] == 1.0)
     else:
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -204,8 +204,10 @@ def test_asarray_copy(library):
     b = asarray(a, copy=None)
     assert is_lib_func(b)
     a[0] = 0.0
-    if library == 'cupy':
+    if library in ('cupy', 'dask.array'):
         # A copy is required for libraries where the default device is not CPU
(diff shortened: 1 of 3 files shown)


ev-br:
Two issues:
- on python 3.9 we install an earlier dask version, which does not have a breaking change
- array-api-tests hang, https://github.com/data-apis/array-api-compat/issues/210

betatim:
In [the standard for `asarray`](https://data-apis.org/array-api/draft/API_specification/generated/array_api.asarray.html#array_api.asarray) it says that for `copy=None` you should not copy if possible. Do you think "in dask we always make a copy" is compatible with that? And if no, is there anything we can do in `array-api-compat` to fix this? It would be nice if we could organise (through the compat layer) for users to get the same behaviour no matter what version of dask they are using.

If there is nothing we can do to fix this then I think updating the test is fine.

lucascolley:
Yes, I think it is compatible for Dask to say that they will always copy, and just raise an error for `copy=False`.

ev-br:
The issue with 2024.12 is `copy=None`, actually.

lucascolley:
> The issue with 2024.12 is copy=None, actually.

It is compatible for them to always copy with `copy=None`.

ev-br:
CI is green with three tweaks
- asarray(..., copy=None) for array-likes always copies
- the test for copy=None is updated to match
- a unrelated test needs to be skipped on CI: was xfailed and was hanging with dask 2024.12

tests/test_common.py
```diff
@@ -204,8 +204,10 @@ def test_asarray_copy(library):
     b = asarray(a, copy=None)
     assert is_lib_func(b)
     a[0] = 0.0
-    if library == 'cupy':
+    if library in ('cupy', 'dask.array'):
         # A copy is required for libraries where the default device is not CPU
+        # dask made a breaking change in 2024.12: copy=None copies
```
crusaderky, line 209:
```suggestion
        # dask changed behaviour of copy=None in 2024.12 to copy
```
IMHO not a breaking change

crusaderky (approved):

lucascolley (approved):

ev-br:
Let's unblock everything then :-)

crusaderky (commented):
See also: https://github.com/data-apis/array-api-compat/pull/235

Issue #223, opened 2024-12-27
Support 2024.12


ev-br:
This is a tracking issue for the 2024.12 revision support in array-api-compat.

The 2024.12 is in a draft still, so this list below is preliminary:

- [x] `diff` : https://github.com/data-apis/array-api-compat/pull/224
- [x] `nextafter` : available in unwrapped namespaces
- [x] `cumulative_prod` : https://github.com/data-apis/array-api-compat/pull/224
- [x] `reciprocal` : available in unwrapped namespaces
- [x] `count_nonzero` : https://github.com/data-apis/array-api-compat/pull/224
- [x] `take_along_axis` : https://github.com/data-apis/array-api-compat/pull/224
- [x] `max_dimensions` in `capabilities` : https://github.com/data-apis/array-api-compat/pull/224
- [x] `mean` is defined for complex arrays : defined in unwrapped namespaces
- [x] `result_type` accepts python scalars : https://github.com/data-apis/array-api-compat/pull/224
.
- [ ] Device-dependent type promotion rules, per https://github.com/data-apis/array-api/pull/889

crusaderky:
take_along_axis in dask is a very complicated matter when there are multiple chunks: https://github.com/dask/dask/issues/3663

mdhaber:
Can we start implementing the draft standard in array API compat, especially `result_type`?

Before scalars were allowed in `array_namespace` and `result type`, I don't think it would be unusual do something like:
```python3
xp = array_namespace(x, y)
x = xp.asarray(x)
y = xp.asarray(y)
dtype = result_type(x, y)
```

especially since the following woudl not be acceptable according to the standard if one of `x` or `y` can be a Python scalar:

```python3
xp = array_namespace(x, y)
dtype = result_type(x, y)  # fails for array_api_strict if `x` or `y` is a Python scalar
```

But if the intent of `result_type` is to find the correct result dtype of an operation between `x` and `y`, the former can produce an incorrect result when scalars are involved, e.g.:
```python3
from array_api_compat import array_namespace, numpy as np
x = np.asarray([1, 2, 3], dtype=np.float32)
y = 3.

xp = array_namespace(x, y)
x_ = np.asarray(x)  # float32
y_ = xp.asarray(y)  # float64

xp.result_type(x_, y_)  # float64
(x*y).dtype  # float32
``` 

[...]

PR #224, opened 2024-12-28
Add draft support for 2024.12 revision

ev-br:
Add missing wrappers from https://github.com/data-apis/array-api-compat/issues/223

None of these is tested, so should be considered as draft.


ev-br:
I'm going to merge this "soon", once the 2024.12 spec is finalized.

PR #227, opened 2025-01-06
TST: run parameterized tests on more libraries

crusaderky:
- Add array-api-strict and sparse to several parameterized tests
- Add array-api-strict to the documentation
- Detect and xfail more bugs in cross-library `asarray()`
- Follow-up: #232 

**Q:**  (out of scope) shouldn't we add an `is_array_api_strict_array` function, to match `is_array_api_strict_namespace`?

PR #228, opened 2025-01-06
ENH: is_lazy_array()

### LINKED ISSUE #225: Conditionally run health checks on jitted JAX arrays and dask arrays
I've found cases of functions in scipy, e.g.

https://github.com/scipy/scipy/blob/4758525cae48f9cfb6971be6702fbb412e783aa5/scipy/cluster/vq.py#L332-L334

that crash on the very first few lines with JAX when they're inside `@jax.jit`:

```python
>>> import jax
>>> import jax.numpy as xp
>>> from scipy.cluster.vq import kmeans
>>> a = xp.asarray([[1.,2.],[3.,4.]])
>>> jax.jit(kmeans)(a, 2)

TracerBoolConversionError: Attempted boolean conversion of traced array with shape bool[].
The error occurred while tracing the function kmeans at /home/crusaderky/github/scipy/build-install/lib/python3.12/site-packages/scipy/cluster/vq.py:332 for jit. This concrete value was not available in Python because it depends on the value of the argument obs.
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.TracerBoolConversionError
```

The issue is that there is a health check enabled by default, `check_finite=True`, which triggers this code:

https://github.com/scipy/scipy/blob/4758525cae48f9cfb6971be6702fbb412e783aa5/scipy/_lib/_array_api.py#L104-L108

A JAX jitted array crashes on `bool()`; a dask array is quietly computed when you do so - which is possibly even worse.

There are two issues here:
1. the default behaviour of the function is to inspect the contents of the array, and
2. the error message is uncomprehensible to an end user, as it is triggered by code deep inside the scipy implementation.

My proposal:

1. in array-api-compat, add two functions:
```python
def is_jax_jitted_array(x):
    return isinstance(x, DynamicJaxprTracer)

def is_material_array(x):
    """Return True if x has contents in memory at the moment of calling this function,
    which are cheap to retrieve as long as they're small in size.
    Return False if x is a future or it would be otherwise impossible or expensive to
    read its contents, regardless of their size.
    """
    return not is_dask_array(x) and not is_jax_jitted_array(x)
```

2. in scipy, change `kmeans(..., check_finite=True)` to `kmeans(..., check_finite=None)`, which will mean "check if possible", and replace
```python
if check_finite:
    _check_finite(x, xp)
```
with
```python
def _check_material_array(x: Array, check: bool | None, check_name: str) -> bool:
    if check is None:
        return is_material_array(x)
    if check and not is_material_array(x):
        raise TypeError(f"Can't check non-material array {type(x)}. Please set {check_name} to None or False.")
    return check

...

if _check_material_array(x, check_finite, "check_finite"):
    _check_finite(x, xp)
```
    
However, @jakevdp mentioned elsewhere that DynamicJaxprTracer is not part of the public API of JAX and there is no public method to test for jitting. Not sure I can see a way forward without this information.

Also CC @rgommers @lucascolley

crusaderky:
Closes #225

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -824,6 +824,63 @@ def is_writeable_array(x) -> bool:
     return True
 
 
+def is_lazy_array(x) -> bool:
+    """Return True if x is potentially a future or it may be otherwise impossible or
+    expensive to eagerly read its contents, regardless of their size, e.g. by
+    calling ``bool(x)`` or ``float(x)``.
+
+    Return False otherwise; e.g. ``bool(x)`` etc. is guaranteed to succeed and to be
+    cheap as long as the array has the right dtype and size.
+
+    Note
+    ----
+    This function errs on the side of caution for array types that may or may not be
+    lazy, e.g. JAX arrays, by always returning True for them.
+    """
+    if (
+        is_numpy_array(x)
+        or is_cupy_array(x)
+        or is_torch_array(x)
+        or is_pydata_sparse_array(x)
+    ):
+        return False
+
+    # **JAX note:** while it is possible to determine if you're inside or outside
+    # jax.jit by testing the subclass of a jax.Array object, as well as testing bool()
+    # as we do below for unknown arrays, this is not recommended by JAX best practices.
+
+    # **Dask note:** Dask eagerly computes the graph on __bool__, __float__, and so on.
+    # This behaviour, while impossible to change without breaking backwards
+    # compatibility, is highly detrimental to performance as the whole graph will end
+    # up being computed multiple times.
+
+    if is_jax_array(x) or is_dask_array(x) or is_ndonnx_array(x):
+        return True
+
+    # Unknown Array API compatible object. Note that this test may have dire consequences
+    # in terms of performance, e.g. for a lazy object that eagerly computes the graph
+    # on __bool__ (dask is one such example, which however is special-cased above).
+
+    # Select a single point of the array
+    s = size(x)
+    if s is None:
+        return True
+    xp = array_namespace(x)
+    if s > 1:
+        x = xp.reshape(x, (-1,))[0]
+    # Cast to dtype=bool and deal with size 0 arrays
+    x = xp.any(x)
+
+    try:
+        bool(x)
+        return False
+    # The Array API standard dictactes that __bool__ should raise TypeError if the
+    # output cannot be defined.
+    # Here we allow for it to raise arbitrary exceptions, e.g. like Dask does.
+    except Exception:
+        return True
+
+
 __all__ = [
     "array_namespace",
     "device",
@@ -845,6 +902,7 @@ __all__ = [
     "is_pydata_sparse_array",
     "is_pydata_sparse_namespace",
     "is_writeable_array",
+    "is_lazy_array",
     "size",
     "to_device",
 ]
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -1,3 +1,10 @@
+import math
+
+import pytest
+import numpy as np
+import array
+from numpy.testing import assert_allclose
+
 from array_api_compat import (  # noqa: F401
     is_numpy_array, is_cupy_array, is_torch_array,
     is_dask_array, is_jax_array, is_pydata_sparse_array,
@@ -6,14 +13,9 @@ from array_api_compat import (  # noqa: F401
 )
 
 from array_api_compat import (
-    device, is_array_api_obj, is_writeable_array, size, to_device
+    device, is_array_api_obj, is_lazy_array, is_writeable_array, size, to_device
 )
 from ._helpers import import_, wrapped_libraries, all_libraries
-
-import pytest
-import numpy as np
-import array
-from numpy.testing import assert_allclose
 
 is_array_functions = {
     'numpy': 'is_numpy_array',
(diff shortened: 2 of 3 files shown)


array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
     return True
 
 
+def is_lazy_array(x) -> bool:
+    """Return True if x is potentially a future or it may be otherwise impossible or
+    expensive to eagerly read its contents, regardless of their size, e.g. by 
+    calling ``bool(x)`` or ``float(x)``.
+
+    Return True otherwise; e.g. ``bool(x)`` etc. is guaranteed to succeed and to be
```
crusaderky, line 827:
```suggestion
    Return False otherwise; e.g. ``bool(x)`` etc. is guaranteed to succeed and to be
```

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
     return True
 
 
+def is_lazy_array(x) -> bool:
+    """Return True if x is potentially a future or it may be otherwise impossible or
+    expensive to eagerly read its contents, regardless of their size, e.g. by 
+    calling ``bool(x)`` or ``float(x)``.
+
+    Return False otherwise; e.g. ``bool(x)`` etc. is guaranteed to succeed and to be
+    cheap as long as the array is the right dtype.
```
crusaderky, line 828:
```suggestion
    cheap as long as the array has the right dtype.
```

tests/test_common.py
```diff
@@ -92,6 +93,70 @@ def test_is_writeable_array_numpy():
     assert not is_writeable_array(x)
 
 
+@pytest.mark.parametrize("library", all_libraries)
```
crusaderky, line 96: https://github.com/data-apis/array-api-compat/pull/227 increases coverage

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
...
+    # This behaviour, while impossible to change without breaking backwards
+    # compatibility, is highly detrimental to performance as the whole graph will end
+    # up being computed multiple times.
+
+    if is_jax_array(x) or is_dask_array(x) or is_ndonnx_array(x):
+        return True
+
+    # Unknown Array API compatible object. Note that this test may have dire consequences
+    # in terms of performance, e.g. for a lazy object that eagerly computes the graph
+    # on __bool__ (dask is one such example, which however is special-cased above).
```
crusaderky, line 857: This is a good argument for replacing everything below this point with a blind `return False`. Please discuss if you'd prefer it that way.

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
...
+
+    Note
+    ----
+    This function errs on the side of caution for array types that may or may not be
+    lazy, e.g. JAX arrays, by always returning True for them.
+    """
+    if (
+        is_numpy_array(x) 
+        or is_cupy_array(x) 
+        or is_torch_array(x)
```
crusaderky, line 838: FIXME is this correct and safe? I don't know enough about torch.compile myself.

tests/test_common.py
```diff
@@ -92,6 +93,75 @@ def test_is_writeable_array_numpy():
...
+    import jax
+
+    x = xp.asarray(array)
+    # Prevent is_lazy_array() from special-casing JAX
+    monkeypatch.setattr(
+        array_api_compat.common._helpers, 
+        "is_jax_array", 
+        lambda x: False,
+    )
+    monkeypatch.setattr(  # Needed for JAX < 0.4.32
```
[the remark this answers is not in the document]
crusaderky: Nevermind, I rewrote the tests to be a lot less hacky

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
...
+        x = xp.reshape(x, (-1,))[0]
+    # Cast to dtype=bool and deal with size 0 arrays
+    x = xp.any(x)
+
+    try:
+        bool(x)
+        return False
+    # The Array API standard dictactes that __bool__ should raise TypeError if the
+    # output cannot be defined.
+    # Here we're more lenient and also allow for e.g. NotImplementedError.
```
crusaderky, line 874:
```suggestion
    # Here we're more lenient and allow for it to raise anything else, e.g. like Dask does.
```

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
...
+        x = xp.reshape(x, (-1,))[0]
+    # Cast to dtype=bool and deal with size 0 arrays
+    x = xp.any(x)
+
+    try:
+        bool(x)
+        return False
+    # The Array API standard dictactes that __bool__ should raise TypeError if the
+    # output cannot be defined.
+    # Here we're more lenient and allow for it to raise anything else, e.g. like Dask does.
```
crusaderky, line 874:
```suggestion
    # Here we allow for it to raise arbitrary exceptions, e.g. like Dask does.
```

tests/test_common.py
```diff
@@ -1,19 +1,22 @@
...
     is_numpy_array, is_cupy_array, is_torch_array,
     is_dask_array, is_jax_array, is_pydata_sparse_array,
     is_numpy_namespace, is_cupy_namespace, is_torch_namespace,
     is_dask_namespace, is_jax_namespace, is_pydata_sparse_namespace,
 )
 
-from array_api_compat import device, is_array_api_obj, is_writeable_array, to_device
-
+from array_api_compat import (
+    device, is_array_api_obj, is_lazy_array, is_writeable_array, to_device, size
```
crusaderky, line 16:
```suggestion
    device, is_array_api_obj, is_lazy_array, is_writeable_array, size, to_device
```

tests/test_common.py
```diff
@@ -92,6 +95,62 @@ def test_is_writeable_array_numpy():
...
+
+
+@pytest.mark.parametrize("exc", [TypeError, AssertionError])
+def test_is_lazy_array_bool_raises(exc, monkeypatch):
+    """Test is_lazy_array() on an unknown Array API compliant object
+    where calling bool() raises:
+    - TypeError: e.g. like jitted JAX. This is the proper exception which
+      lazy arrays should raise as per the Array API specification
+    - something else: e.g. like Dask, where bool() triggers compute()
+      which can cause in any kind of exception to be raised
```
crusaderky, line 125:
```suggestion
      which can result in any kind of exception to be raised
```

tests/test_common.py
```diff
@@ -92,6 +95,45 @@ def test_is_writeable_array_numpy():
...
+def test_is_lazy_array(library):
+    lib = import_(library)
+    x = lib.asarray([1, 2, 3])
+    assert isinstance(is_lazy_array(x), bool)
+
+
+@pytest.mark.parametrize("shape", [(math.nan,), (1, math.nan), (None, ), (1, None)])
+def test_is_lazy_array_nan_size(shape, monkeypatch):
+    """Test is_lazy_array() on an unknown Array API compliant object
+    with NaN (like Dask) or None (like ndonnx) in its shape
```
crusaderky, line 108: To be simplified after https://github.com/data-apis/array-api-compat/pull/231

array_api_compat/common/_helpers.py
```diff
@@ -819,6 +819,63 @@ def is_writeable_array(x) -> bool:
...
+    if is_jax_array(x) or is_dask_array(x) or is_ndonnx_array(x):
+        return True
+
+    # Unknown Array API compatible object. Note that this test may have dire consequences
+    # in terms of performance, e.g. for a lazy object that eagerly computes the graph
+    # on __bool__ (dask is one such example, which however is special-cased above).
+
+    # Select a single point of the array
+    s = size(x)
+    if s is None or math.isnan(s):
```
crusaderky, line 861: To be simplified after https://github.com/data-apis/array-api-compat/pull/231

array_api_compat/common/_helpers.py
rgommers: It works, at least for most backends, since TorchDynamo can handle graph breaks. If we have to make a choice here, then I'd say what you wrote is correct for PyTorch.

rgommers (approved):
LGTM, thanks @crusaderky. And `is_lazy_array` seems okay as a helper function in this library rather than in `array-api-extra`.

@ev-br @asmeurer @lucascolley any of you want to take a look?

lucascolley (approved):

ev-br (approved):
Let's roll with this indeed, if only to unblock the jax work. FWIW I'm not sure about the "unknown library" dance, it looks complicated and potentially costly (as discussed in the comments), but that's something we'll be able to adjust en route. Thanks @crusaderky, all.

PR #231, opened 2025-01-07
ENH: size() to return None on dask instead of nan

crusaderky:
dask arrays of unknown shape return NaN in their `.size` and `.shape` properties. This was chosen a long time before the Array API standard existed, but alas diverges from the array API standard itself:

- https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.size.html
- https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.shape.html

ndonnx, another library which can have arrays of unknown shape, uses None instead.

This PR changes the `size()` function to make dask conform to the standard and adds units tests which were previously missing.

**Q:** Should we add a `shape()` function to cover for Dask's `.shape` property non-standard behaviour?

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -5,8 +5,9 @@ from array_api_compat import (  # noqa: F401
     is_dask_namespace, is_jax_namespace, is_pydata_sparse_namespace,
 )
 
-from array_api_compat import device, is_array_api_obj, is_writeable_array, to_device
-
+from array_api_compat import (
+    device, is_array_api_obj, is_writeable_array, size, to_device
+)
 from ._helpers import import_, wrapped_libraries, all_libraries
 
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_common.py
```diff
@@ -92,6 +93,28 @@ def test_is_writeable_array_numpy():
...
 
 
+@pytest.mark.parametrize("library", all_libraries)
+def test_size(library):
+    xp = import_(library)
+    x = xp.asarray([1, 2, 3])
+    assert size(x) == 3
+
+
+@pytest.mark.parametrize("library", all_libraries)
```
crusaderky, line 103: FIXME this doesn't actually test ndonnx's None case yet!

I need https://github.com/data-apis/array-api-compat/pull/227 to be merged before I can open a new PR specific to ndonnx.

crusaderky:
CI failure seems unrelated

ev-br (approved):
LGTM, thank you @crusaderky

PR #232, opened 2025-01-08
ENH: ndonnx device() support; TST: better ndonnx test coverage

crusaderky:
- Add support to ndonnx in device() and to_device().
- Add ndonnx to parameterized tests
- Add ndonnx to the documentation
- Adds key use cases to #228
- Adds key use cases to #231

### DIFF
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -645,7 +645,7 @@ def device(x: Array, /) -> Device:
     to_device : Move array data to a different device.
 
     """
-    if is_numpy_array(x):
+    if is_numpy_array(x) or is_ndonnx_array(x):
         return "cpu"
     elif is_dask_array(x):
         # Peek at the metadata of the Dask array to determine type
@@ -776,7 +776,7 @@ def to_device(x: Array, device: Device, /, *, stream: Optional[Union[int, Any]] 
     device : Hardware device the array data resides on.
 
     """
-    if is_numpy_array(x):
+    if is_numpy_array(x) or is_ndonnx_array(x):
         if stream is not None:
             raise ValueError("The stream argument to to_device() is not supported")
         if device == 'cpu':
--- a/array_api_compat/common/_helpers.py
+++ b/array_api_compat/common/_helpers.py
@@ -645,7 +645,7 @@ def device(x: Array, /) -> Device:
     to_device : Move array data to a different device.
 
     """
-    if is_numpy_array(x) or is_ndonnx_array(x):
+    if is_numpy_array(x):
         return "cpu"
     elif is_dask_array(x):
         # Peek at the metadata of the Dask array to determine type
--- a/tests/_helpers.py
+++ b/tests/_helpers.py
@@ -3,11 +3,12 @@ from importlib import import_module
 import pytest
 
 wrapped_libraries = ["numpy", "cupy", "torch", "dask.array"]
-all_libraries = wrapped_libraries + ["array_api_strict", "jax.numpy", "sparse"]
-
+all_libraries = wrapped_libraries + [
+    "array_api_strict", "jax.numpy", "ndonnx", "sparse"
+]
 
 def import_(library, wrapper=False):
-    if library == 'cupy':
+    if library in ('cupy', 'ndonnx'):
         pytest.importorskip(library)
     if wrapper:
         if 'jax' in library:
--- a/tests/_helpers.py
+++ b/tests/_helpers.py
@@ -21,3 +21,14 @@ def import_(library, wrapper=False):
             library = 'array_api_compat.' + library
 
     return import_module(library)
+
+
+def xfail(request: pytest.FixtureRequest, reason: str) -> None:
+    """
+    XFAIL the currently running test.
+
+    Unlike ``pytest.xfail``, allow rest of test to execute instead of immediately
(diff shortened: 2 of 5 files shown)


crusaderky:
CI failure is unrelated

array_api_compat/common/_helpers.py
```diff
@@ -772,7 +772,7 @@ def to_device(x: Array, device: Device, /, *, stream: Optional[Union[int, Any]]
     device : Hardware device the array data resides on.
 
     """
-    if is_numpy_array(x):
+    if is_numpy_array(x) or is_ndonnx_array(x):
```
ev-br, line 775: What I don't quite understand is this: https://github.com/data-apis/array-api-compat/pull/232/files#diff-d5d8fa69860e03769cc235a54027efad4376a8c48eeab71aaa11365ea308123bR141 states that the array API support is internal to `ndonnx`, which seems to imply there's no need to special-case it here?

The note reads "Similar to JAX,  `ndonnx` Array API support is contained directly in `ndonnx`." --- and indeed, there are no workarounds for jax in the codebase.

tests/_helpers.py
```diff
@@ -3,11 +3,12 @@
 import pytest
 
 wrapped_libraries = ["numpy", "cupy", "torch", "dask.array"]
-all_libraries = wrapped_libraries + ["array_api_strict", "jax.numpy", "sparse"]
-
+all_libraries = wrapped_libraries + [
+    "array_api_strict", "jax.numpy", "ndonnx", "sparse"
```
ev-br, line 7: This too. Why are we testing here libraries which we don't wrap?

ev-br (changes requested):
What concerns me here is a sort of conceptual question: What is the scope of a test run here in `array-api-compat`? 

Basically, there are two kinds of array libraries: those which are array API compatible and those which need some wrapping. `array-api-compat` does the latter, for those libraries which need it. 

Than why are we testing the libraries which we do not wrap?

There's no question that we need some better ergonomics for running tests across array libraries. I'm just not convinced it's a good idea to sort of sneak this testing here. Let's discuss a structural solution instead.

crusaderky:
> Then why are we testing the libraries which we do not wrap?

because the libraries that we do not wrap still need the functions defined in `_helpers` to work around their quirks.

array_api_compat/common/_helpers.py
crusaderky: Because they support the standard imperfectly. The array-api-tests for `.device` and `.to_device` are XFAILed:
https://github.com/Quantco/ndonnx/actions/workflows/array-api.yml

```
api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_method-to_device] 
[gw1] [ 13%] XFAIL api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_method-to_device] 
api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_attribute-T] 
[gw1] [ 14%] PASSED api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_attribute-T] 
api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_attribute-device] 
[gw1] [ 14%] XFAIL api-coverage-tests/array_api_tests/test_has_names.py::test_has_names[array_attribute-device] 
```

Adding `.device` and `.to_device` won't break _any_ library's backwards compatibility, so it would be best if each library just fixed it on their side. And yet here we have a function that works around the quirks of every library - last but not least to support least-than-newest versions of it.

tests/_helpers.py
crusaderky: Because `_helpers.py` is chock full of special cases and exceptions for them.

crusaderky:
CC @adityagoel4512

array_api_compat/common/_helpers.py
adityagoel4512: I will address this upstream. We may have to do something similar to what you're doing with Dask since it's not necessarily true that the computation occurs on the CPU. In fact, it's not strictly in our control since a serialized ONNX graph from ndonnx could later be executed on many different backends.

adityagoel4512: I will address this upstream. I should say that it's not necessary that the computation occurs on CPU. In fact, it's not strictly in our control which device the array lives on since a serialized ONNX graph from ndonnx could later be executed on many different hardware targets or runtimes, so we're one level removed from the execution.

Just from a cursory glance of the device page of the specification, we may do something similar to what you're doing with Dask.

adityagoel4512:
@crusaderky we added device support in https://github.com/Quantco/ndonnx/commit/4ed6af06501fd75b354ddb17fcd8c9ef6b6af9f9. It should be present in the next release. Is it urgent that we land this PR soon?

crusaderky:
> @crusaderky we added device support in [Quantco/ndonnx@039264d](https://github.com/Quantco/ndonnx/commit/039264da5587edc5a65ebc6ac2f7f0d0f19176de). It should be possible to remove the device support here once we make a release. Is it urgent that we land this PR soon?

No, this PR can wait.

crusaderky:
I've reverted the device support and xfailed the device test for now.
This PR is ready for final review and merge.

ev-br (approved):

ev-br:
Okay, let's roll with it. Thank you @crusaderky

PR #235, opened 2025-01-15
`da.asarray` should not materialize the graph

crusaderky:


array_api_compat/dask/array/_aliases.py
```diff
@@ -40,19 +40,25 @@
 isdtype = get_xp(np)(_aliases.isdtype)
 unstack = get_xp(da)(_aliases.unstack)
 
+# da.astype doesn't respect copy=True
```
crusaderky, line 43: All changes to functions other than  asarray are purely cosmetic

array_api_compat/dask/array/_aliases.py
```diff
@@ -40,19 +40,25 @@
...
 def astype(
     x: Array,
     dtype: Dtype,
     /,
     *,
     copy: bool = True,
-    device: Device | None = None
+    device: Optional[Device] = None
```
crusaderky, line 50: this change only matters if someone runs mypy/pyright in python 3.9

array_api_compat/dask/array/_aliases.py
```diff
@@ -83,13 +93,12 @@ def _dask_arange(
         # prepend the default value for start which is 0
         args.insert(0, 0)
     args.append(step)
-    return xp.arange(*args, dtype=dtype, **kwargs)
 
-arange = get_xp(da)(_dask_arange)
-eye = get_xp(da)(_aliases.eye)
```
crusaderky, line 89: duplicate

crusaderky:
@ev-br this is ready for review and merge

ev-br (commented):
The testing pattern is a bit bizzare, I've to admit. Would be nice if there was a way to make a little more "boring" and explicit.

crusaderky:
> The testing pattern is a bit bizzare, I've to admit. Would be nice if there was a way to make a little more "boring" and explicit.

I've replaced the fixture with a context manager to make it clearer.

ev-br (approved):
Thanks @crusaderky !

PR #237, opened 2025-01-17
ENH: is_lazy_array and is_writeable_array to return False on non-arrays

crusaderky:


array_api_compat/common/_helpers.py
```diff
@@ -254,10 +262,12 @@ def is_array_api_obj(x):
...
         or hasattr(x, '__array_namespace__')
 
-def _compat_module_name():
+
+def _compat_module_name() -> str:
     assert __name__.endswith('.common._helpers')
     return __name__.removesuffix('.common._helpers')
 
+
 def is_numpy_namespace(xp) -> bool:
```
crusaderky, line 271: Sphinx gave me grief over `ModuleType`, so I left it out of scope for this PR.

ev-br:
LGTM, thanks @crusaderky

PR #238, opened 2025-01-21
BUG: `.device` attribute inside `jax.jit`

crusaderky:
- Work around https://github.com/jax-ml/jax/issues/26000 on all versions of JAX
- Fix `to_device` on JAX < 0.4.31 inside jax.jit

### DIFF
--- a/tests/test_jax.py
+++ b/tests/test_jax.py
@@ -1,0 +1,34 @@
+import jax
+import jax.numpy as jnp
+from numpy.testing import assert_equal
+import pytest
+
+from array_api_compat import device, to_device
+
+HAS_JAX_0_4_31 = jax.__version__ >= "0.4.31"
+
+
+@pytest.mark.parametrize(
+    "func", 
+    [
+        lambda x: jnp.zeros(1, device=device(x)),
+        lambda x: jnp.zeros_like(jnp.ones(1, device=device(x))),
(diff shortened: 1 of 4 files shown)


tests/test_jax.py
```diff
@@ -0,0 +1,34 @@
...
+    ]
+)
+def test_device_jit(func):
+    # Test work around to https://github.com/jax-ml/jax/issues/26000
+    # Also test missing to_device() method in JAX < 0.4.31
+    # when inside jax.jit, even after importing jax.experimental.array_api
+
+    x = jnp.ones(1)
+    assert_equal(func(x), jnp.asarray([0]))
+    assert_equal(jax.jit(func)(x), jnp.asarray([0]))
```
lucascolley, line 34: is the idea that, if the workaround was incorrect, `assert_equal` would choke trying to compare devices on different devices?

crusaderky: no, assert_equal doesn't compare devices. This would choke well before returning from `func`.

crusaderky: If you have multiple devices and x is on a non-default one, the output of func will be on the wrong device.
At any rate, CI hosts have only one jax device, so we don't have the means to test this.

lucascolley (approved):
LGTM, thanks @crusaderky !

PR #239, opened 2025-01-23
ENH: Dask: `sort` and `argsort`

crusaderky:
Crude implementation of `sort` and `argsort` for dask.array, which is functionally correct but can be extremely memory and network-intensive.

A better solution would be to implement these two functions in dask.array itself, on top of the shuffle subsystem which is already used for `dask.dataframe.DataFrame.sort_values`.

FYI @fjetter @phofl @hendrikmakait

### DIFF
--- a/array_api_compat/dask/array/_aliases.py
+++ b/array_api_compat/dask/array/_aliases.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
-from ...common import _aliases
+from typing import Callable
+
+from ...common import _aliases, array_namespace
 
 from ..._internal import get_xp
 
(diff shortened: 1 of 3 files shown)


array_api_compat/dask/array/_aliases.py
```diff
@@ -228,10 +230,84 @@ def _isscalar(a):
...
+    TODO Implement sort and argsort properly in Dask on top of the shuffle subsystem.
+    """
+    if axis < 0:
+        axis += x.ndim
+    rechunk = False
+    if x.numblocks[axis] > 1:
+        rechunk = True
+        # Break chunks on other axes in an attempt to keep chunk size low
+        x = x.rechunk({i: -1 if i == axis else "auto" for i in range(x.ndim)})
+    meta_xp = array_namespace(x._meta)
```
crusaderky, line 289: FIXME: I couldn't write a test that actually uses a meta_xp other than numpy, because `da.from_array(array_api_strict.asarray(0))` doesn't work in Dask yet.

lucascolley (commented):
would you like to get feedback from a Dask expert before we merge this? Or are you confident that it is at least good enough for now?

crusaderky:
I'm confident that this is tolerable at least as a temporary crutch. It will work for some geometries and will go OOM for others, which IMHO is vastly better than not having anything at all.

lucascolley (approved):

PR #244, opened 2025-01-29
ENH: More uint types for torch

crusaderky:


ev-br:
Note that due to the way automation is set up in this repository, Ci failures --- even if unrelated --- block merges and releases (hint hint).
- `clip` failure is a flaky test, might disappear on a rerun (a better fix would be much appreciated of course)
- dask's error, `Original Message: Can not use auto rechunking with object dtype. We are unable to estimate the size in bytes of object data`, is something new and probably due to recent dask effort

crusaderky:
https://github.com/data-apis/array-api-tests/issues/338

ev-br:
The change LGTM. It effectively bumps the torch requirement to >= 2.3, which is likely fine. 

I'm going to carpe diem while the CI is green after https://github.com/data-apis/array-api-compat/pull/246 and press the green button. 

Thanks @crusaderky

PR #246, opened 2025-01-30
TST: da.arange fails for args near 2**63

crusaderky:
https://github.com/dask/dask/issues/11706

Closes https://github.com/data-apis/array-api-tests/issues/338

Also clean up a wealth of XPASS


crusaderky:
@ev-br ready for review

ev-br (approved):
A cleanup of dark xfails is both very welcome and scary: some (many?) of these can be flaky tests. That said, undoing them en masse and re enabling those which do show up is a valid strategy, let's do this. 
And we do need an xfail for the arange test which blocks the world. 
Thanks @crusaderky

PR #251, opened 2025-02-03
WIP: add a CuPy CI run

### LINKED ISSUE #197: Run CuPy tests on CI
We need to run the CuPy tests on CI somehow. Right now, [I have to manually run CuPy tests locally when I do a release](https://data-apis.org/array-api-compat/dev/releasing.html). But this only works because I have a local machine that has a CUDA GPU. If someone else wanted to do a release, they would need to access such a machine. But also, this means that we only really check CuPy when we do a release. I also check it manually when doing a CuPy related PR, but having it run more often would be nice. It's also easy to forget to do this when releasing. 

There are ways of getting GPU runners for CI, which we should use here.

ev-br:
closes gh-197


rgommers:
Thanks for starting on this @ev-br. This is mostly right, but it should use Pixi like we do on SciPy, rather than pull CuPy from PyPI - that is good for both maintenance load (same config) and for stability & performance. CUDA libraries, compilers, etc. all come from conda-forge. And that is also better for the ability to run the CI job locally with the exact same dependencies.

I'll figure out how to connect the admin pieces (credit card & billing).

ev-br:
This is too stale to be useful. Closing

PR #252, opened 2025-02-20
Revert change to ndonnx .to_device

crusaderky:
Revert change that slipped into the final version of #232 from its first draft


ev-br:
Thanks Guido

PR #255, opened 2025-02-20
Add pyproject.toml

NeilGirdhar:

PR #254, opened 2025-02-20
Annotate array_namespace and add py.typed

NeilGirdhar:
This appears to be the only missing type annotation in the project.

PR #250, opened 2025-02-02
release 1.11 branch

ev-br:
TODO:

- [x] wait until 2024.12 is actually a thing
- [x] merge outstanding PRs (depends on what ends up in 2024.12 revision)
- [x] update the changelog
- [x] update the array API version everywhere
- [x] clarify the torch uint status, viz https://github.com/data-apis/array-api-compat/pull/253
- [x] back out `py.typed`
- [x] bump the version

PR #257, opened 2025-02-27
TYP: Type annotations overhaul, part 1

crusaderky:
Review all type annotations.

- dtypes were called a mix of `DType` and `Dtype`. I picked `DType` in accordance with `numpy.typing`.
- arrays were called a mix of `Array`, `ndarray`, or `array`. I picked `Array` as it is more idiomatic.
- Added Namespace annotation
- Removed many `TYPE_CHECKING` guards, which were doing nothing but reduce readability (performance impact at runtime is absolutely inconsequential)
- Add `py.typed`. Note that this will typically not work for `array_api_compat.<namespace>.` modules as they are typically pulled in by `array_namespace`, so for most people it is exclusively about `array_api_compat.common._helpers`.

#### Out of scope
- Add type checkers to CI
- Make `common._typing` public. This will be in a later PR.

### DIFF
--- a/array_api_compat/common/_aliases.py
+++ b/array_api_compat/common/_aliases.py
@@ -4,15 +4,11 @@ These are functions that are just aliases of existing functions in NumPy.
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-if TYPE_CHECKING:
-    from typing import Optional, Sequence, Tuple, Union
-    from ._typing import ndarray, Device, Dtype
-
-from typing import NamedTuple
 import inspect
+from typing import Any, NamedTuple, Optional, Sequence, Tuple, Union
 
 from ._helpers import array_namespace, _check_device, device, is_torch_array, is_cupy_namespace
+from ._typing import Array, Device, DType, Namespace
 
 # These functions are modified from the NumPy versions.
 
@@ -24,29 +20,34 @@ def arange(
     stop: Optional[Union[int, float]] = None,
     step: Union[int, float] = 1,
     *,
-    xp,
-    dtype: Optional[Dtype] = None,
+    xp: Namespace,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
-    **kwargs
-) -> ndarray:
+    **kwargs,
+) -> Array:
     _check_device(xp, device)
     return xp.arange(start, stop=stop, step=step, dtype=dtype, **kwargs)
 
 def empty(
     shape: Union[int, Tuple[int, ...]],
-    xp,
+    xp: Namespace,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
-    **kwargs
-) -> ndarray:
+    **kwargs,
+) -> Array:
     _check_device(xp, device)
     return xp.empty(shape, dtype=dtype, **kwargs)
 
 def empty_like(
-    x: ndarray, /, xp, *, dtype: Optional[Dtype] = None, device: Optional[Device] = None,
-    **kwargs
-) -> ndarray:
+    x: Array,
+    /,
+    xp: Namespace,
+    *,
+    dtype: Optional[DType] = None, 
+    device: Optional[Device] = None,
+    **kwargs,
+) -> Array:
     _check_device(xp, device)
     return xp.empty_like(x, dtype=dtype, **kwargs)
 
@@ -55,37 +56,37 @@ def eye(
     n_cols: Optional[int] = None,
     /,
     *,
-    xp,
+    xp: Namespace,
     k: int = 0,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.eye(n_rows, M=n_cols, k=k, dtype=dtype, **kwargs)
 
 def full(
     shape: Union[int, Tuple[int, ...]],
     fill_value: Union[int, float],
-    xp,
+    xp: Namespace,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.full(shape, fill_value, dtype=dtype, **kwargs)
 
 def full_like(
-    x: ndarray,
+    x: Array,
     /,
     fill_value: Union[int, float],
     *,
-    xp,
-    dtype: Optional[Dtype] = None,
+    xp: Namespace,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.full_like(x, fill_value, dtype=dtype, **kwargs)
 
@@ -95,48 +96,58 @@ def linspace(
     /,
     num: int,
     *,
-    xp,
-    dtype: Optional[Dtype] = None,
+    xp: Namespace,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     endpoint: bool = True,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.linspace(start, stop, num, dtype=dtype, endpoint=endpoint, **kwargs)
 
 def ones(
     shape: Union[int, Tuple[int, ...]],
-    xp,
+    xp: Namespace,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.ones(shape, dtype=dtype, **kwargs)
 
 def ones_like(
-    x: ndarray, /, xp, *, dtype: Optional[Dtype] = None, device: Optional[Device] = None,
+    x: Array,
+    /,
+    xp: Namespace,
+    *,
+    dtype: Optional[DType] = None,
+    device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.ones_like(x, dtype=dtype, **kwargs)
 
 def zeros(
     shape: Union[int, Tuple[int, ...]],
-    xp,
+    xp: Namespace,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.zeros(shape, dtype=dtype, **kwargs)
 
 def zeros_like(
-    x: ndarray, /, xp, *, dtype: Optional[Dtype] = None, device: Optional[Device] = None,
+    x: Array,
+    /,
+    xp: Namespace,
+    *,
+    dtype: Optional[DType] = None,
+    device: Optional[Device] = None,
     **kwargs,
-) -> ndarray:
+) -> Array:
     _check_device(xp, device)
     return xp.zeros_like(x, dtype=dtype, **kwargs)
 
@@ -150,23 +161,23 @@ def zeros_like(
 # Note that these named tuples aren't actually part of the standard namespace,
 # but I don't see any issue with exporting the names here regardless.
 class UniqueAllResult(NamedTuple):
-    values: ndarray
-    indices: ndarray
-    inverse_indices: ndarray
-    counts: ndarray
+    values: Array
+    indices: Array
+    inverse_indices: Array
+    counts: Array
 
 
 class UniqueCountsResult(NamedTuple):
-    values: ndarray
-    counts: ndarray
+    values: Array
+    counts: Array
 
 
 class UniqueInverseResult(NamedTuple):
-    values: ndarray
-    inverse_indices: ndarray
+    values: Array
+    inverse_indices: Array
 
 
-def _unique_kwargs(xp):
+def _unique_kwargs(xp: Namespace) -> dict[str, Any]:
     # Older versions of NumPy and CuPy do not have equal_nan. Rather than
     # trying to parse version numbers, just check if equal_nan is in the
     # signature.
@@ -175,7 +186,7 @@ def _unique_kwargs(xp):
         return {'equal_nan': False}
     return {}
 
-def unique_all(x: ndarray, /, xp) -> UniqueAllResult:
+def unique_all(x: Array, /, xp: Namespace) -> UniqueAllResult:
     kwargs = _unique_kwargs(xp)
     values, indices, inverse_indices, counts = xp.unique(
         x,
@@ -195,7 +206,7 @@ def unique_all(x: ndarray, /, xp) -> UniqueAllResult:
     )
 
 
-def unique_counts(x: ndarray, /, xp) -> UniqueCountsResult:
+def unique_counts(x: Array, /, xp: Namespace) -> UniqueCountsResult:
     kwargs = _unique_kwargs(xp)
     res = xp.unique(
         x,
@@ -208,7 +219,7 @@ def unique_counts(x: ndarray, /, xp) -> UniqueCountsResult:
     return UniqueCountsResult(*res)
 
 
-def unique_inverse(x: ndarray, /, xp) -> UniqueInverseResult:
+def unique_inverse(x: Array, /, xp: Namespace) -> UniqueInverseResult:
     kwargs = _unique_kwargs(xp)
     values, inverse_indices = xp.unique(
         x,
@@ -223,7 +234,7 @@ def unique_inverse(x: ndarray, /, xp) -> UniqueInverseResult:
     return UniqueInverseResult(values, inverse_indices)
 
 
-def unique_values(x: ndarray, /, xp) -> ndarray:
+def unique_values(x: Array, /, xp: Namespace) -> Array:
     kwargs = _unique_kwargs(xp)
     return xp.unique(
         x,
@@ -236,42 +247,42 @@ def unique_values(x: ndarray, /, xp) -> ndarray:
 # These functions have different keyword argument names
 
 def std(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     axis: Optional[Union[int, Tuple[int, ...]]] = None,
-    correction: Union[int, float] = 0.0, # correction instead of ddof
+    correction: Union[int, float] = 0.0,  # correction instead of ddof
     keepdims: bool = False,
     **kwargs,
-) -> ndarray:
+) -> Array:
     return xp.std(x, axis=axis, ddof=correction, keepdims=keepdims, **kwargs)
 
 def var(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     axis: Optional[Union[int, Tuple[int, ...]]] = None,
-    correction: Union[int, float] = 0.0, # correction instead of ddof
+    correction: Union[int, float] = 0.0,  # correction instead of ddof
     keepdims: bool = False,
     **kwargs,
-) -> ndarray:
+) -> Array:
     return xp.var(x, axis=axis, ddof=correction, keepdims=keepdims, **kwargs)
 
 # cumulative_sum is renamed from cumsum, and adds the include_initial keyword
 # argument
 
 def cumulative_sum(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     axis: Optional[int] = None,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     include_initial: bool = False,
-    **kwargs
-) -> ndarray:
+    **kwargs,
+) -> Array:
     wrapped_xp = array_namespace(x)
 
     # TODO: The standard is not clear about what should happen when x.ndim == 0.
@@ -294,15 +305,15 @@ def cumulative_sum(
 
 
 def cumulative_prod(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     axis: Optional[int] = None,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     include_initial: bool = False,
-    **kwargs
-) -> ndarray:
+    **kwargs,
+) -> Array:
     wrapped_xp = array_namespace(x)
 
     if axis is None:
@@ -325,17 +336,18 @@ def cumulative_prod(
 # The min and max argument names in clip are different and not optional in numpy, and type
 # promotion behavior is different.
 def clip(
-    x: ndarray,
+    x: Array,
     /,
-    min: Optional[Union[int, float, ndarray]] = None,
-    max: Optional[Union[int, float, ndarray]] = None,
+    min: Optional[Union[int, float, Array]] = None,
+    max: Optional[Union[int, float, Array]] = None,
     *,
-    xp,
+    xp: Namespace,
     # TODO: np.clip has other ufunc kwargs
-    out: Optional[ndarray] = None,
-) -> ndarray:
+    out: Optional[Array] = None,
+) -> Array:
     def _isscalar(a):
         return isinstance(a, (int, float, type(None)))
+
     min_shape = () if _isscalar(min) else min.shape
     max_shape = () if _isscalar(max) else max.shape
 
@@ -389,15 +401,18 @@ def clip(
     return out[()]
 
 # Unlike transpose(), the axes argument to permute_dims() is required.
-def permute_dims(x: ndarray, /, axes: Tuple[int, ...], xp) -> ndarray:
+def permute_dims(x: Array, /, axes: Tuple[int, ...], xp: Namespace) -> Array:
     return xp.transpose(x, axes)
 
 # np.reshape calls the keyword argument 'newshape' instead of 'shape'
-def reshape(x: ndarray,
-            /,
-            shape: Tuple[int, ...],
-            xp, copy: Optional[bool] = None,
-            **kwargs) -> ndarray:
+def reshape(
+    x: Array,
+    /,
+    shape: Tuple[int, ...],
+    xp: Namespace,
+    copy: Optional[bool] = None,
+    **kwargs,
+) -> Array:
     if copy is True:
         x = x.copy()
     elif copy is False:
@@ -409,9 +424,15 @@ def reshape(x: ndarray,
 # The descending keyword is new in sort and argsort, and 'kind' replaced with
 # 'stable'
 def argsort(
-    x: ndarray, /, xp, *, axis: int = -1, descending: bool = False, stable: bool = True,
+    x: Array,
+    /,
+    xp: Namespace,
+    *,
+    axis: int = -1,
+    descending: bool = False,
+    stable: bool = True,
     **kwargs,
-) -> ndarray:
+) -> Array:
     # Note: this keyword argument is different, and the default is different.
     # We set it in kwargs like this because numpy.sort uses kind='quicksort'
     # as the default whereas cupy.sort uses kind=None.
@@ -434,9 +455,15 @@ def argsort(
     return res
 
 def sort(
-    x: ndarray, /, xp, *, axis: int = -1, descending: bool = False, stable: bool = True,
+    x: Array,
+    /,
+    xp: Namespace,
+    *,
+    axis: int = -1,
+    descending: bool = False,
+    stable: bool = True,
     **kwargs,
-) -> ndarray:
+) -> Array:
     # Note: this keyword argument is different, and the default is different.
     # We set it in kwargs like this because numpy.sort uses kind='quicksort'
     # as the default whereas cupy.sort uses kind=None.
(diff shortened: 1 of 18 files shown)


ev-br:
PSA: CI failing left and right is a transient artifact, will get back to reasonable after gh-250 is in and 1.11 is out.

This is too large to sneak into 1.11 unless we want to delay it (I don't). So this PR is the first large feature of 1.12.

crusaderky:
> This is too large to sneak into 1.11 unless we want to delay it (I don't). So this PR is the first large feature of 1.12.

Then we should temporarily remove `py.typed` introduced in #254 CC @NeilGirdhar

ev-br:
> Then we should temporarily remove py.typed

Yes. I'll do this in gh-250

crusaderky:
# Discussion
The Array API documentation calls the types `array`, `dtype`, and `device` all lowercase.
I'm not a fan for obvious reasons. Should we change them here to be all lowercase, or should we propose changing them to uppercase in array-api?

ev-br:
> The Array API documentation calls the types array, dtype, and device all lowercase. ... Should we change them here to be all lowercase, or should we propose changing them to uppercase in array-api?

Great idea, would be best to coordinate with the spec indeed. Consortium meetings are weekly on Mondays (hint hint).

ev-br:
Also would be best to coordinate with the [array-api-typing](https://github.com/data-apis/array-api-typing) initiative, cc
@jorenham @nstarman

array_api_compat/common/_aliases.py
```diff
@@ -510,8 +537,11 @@ def vecdot(x1: ndarray, x2: ndarray, /, xp, *, axis: int = -1) -> ndarray:
 # isdtype is a new function in the 2022.12 array API specification.
 
 def isdtype(
-    dtype: Dtype, kind: Union[Dtype, str, Tuple[Union[Dtype, str], ...]], xp,
-    *, _tuple=True, # Disallow nested tuples
+    dtype: DType,
+    kind: Union[DType, str, Tuple[Union[DType, str], ...]],
```
jorenham, line 541: You could also use a `typing.Literal` to further tighten this up

```suggestion
    kind: DType | str | tuple[DType | str, ...],
```

array_api_compat/common/_fft.py
```diff
@@ -1,149 +1,148 @@
...
 def fft(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     n: Optional[int] = None,
     axis: int = -1,
     norm: Literal["backward", "ortho", "forward"] = "backward",
```
jorenham, line 18: This literal is repeated quite often, so it's probably a good idea to extract it as something like 

```py
_Norm: TypeAlias = Literal["backward", "ortho", "forward"]
```

array_api_compat/common/_fft.py
```diff
@@ -1,149 +1,148 @@
...
 
 def fftn(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     s: Sequence[int] = None,
     axes: Sequence[int] = None,
```
jorenham, line 45:
```suggestion
    s: Sequence[int] | None = None,
    axes: Sequence[int] | None = None,
```

FYI, a `Sequence[int]` is very broad, and will even accept things like `bytes`. Maybe a `tuple[int, ...]` is more appropriate here.

array_api_compat/common/_fft.py
```diff
@@ -1,149 +1,148 @@
...
 
 def ifftn(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     s: Sequence[int] = None,
     axes: Sequence[int] = None,
```
jorenham, line 59:
```suggestion
    s: Sequence[int] | None = None,
    axes: Sequence[int] | None = None,
```

array_api_compat/common/_fft.py
```diff
@@ -1,149 +1,148 @@
...
 
 def rfftn(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     s: Sequence[int] = None,
     axes: Sequence[int] = None,
```
jorenham, line 101:
```suggestion
    s: Sequence[int] | None = None,
    axes: Sequence[int] | None = None,
```

array_api_compat/common/_fft.py
```diff
@@ -1,149 +1,148 @@
...
 
 def irfftn(
-    x: ndarray,
+    x: Array,
     /,
-    xp,
+    xp: Namespace,
     *,
     s: Sequence[int] = None,
     axes: Sequence[int] = None,
```
jorenham, line 115:
```suggestion
    s: Sequence[int] | None = None,
    axes: Sequence[int] | None = None,
```

array_api_compat/common/_fft.py
```diff
@@ -168,23 +167,27 @@ def fftfreq(
...
     if device not in ["cpu", None]:
         raise ValueError(f"Unsupported device {device!r}")
     res = xp.fft.rfftfreq(n, d=d)
     if dtype is not None:
         return res.astype(dtype)
     return res
 
-def fftshift(x: ndarray, /, xp, *, axes: Union[int, Sequence[int]] = None) -> ndarray:
+def fftshift(
+    x: Array, /, xp: Namespace, *, axes: Union[int, Sequence[int]] = None
```
jorenham, line 184:
```suggestion
    x: Array, /, xp: Namespace, *, axes: int | Sequence[int] | None = None
```

array_api_compat/common/_fft.py
```diff
@@ -168,23 +167,27 @@ def fftfreq(
...
+) -> Array:
     return xp.fft.fftshift(x, axes=axes)
 
-def ifftshift(x: ndarray, /, xp, *, axes: Union[int, Sequence[int]] = None) -> ndarray:
+def ifftshift(
+    x: Array, /, xp: Namespace, *, axes: Union[int, Sequence[int]] = None
```
jorenham, line 189:
```suggestion
    x: Array, /, xp: Namespace, *, axes: int | Sequence[int] | None = None
```

array_api_compat/common/_helpers.py
```diff
@@ -439,7 +437,11 @@ def _check_api_version(api_version: str) -> None:
         raise ValueError("Only the 2024.12 version of the array API specification is currently supported")
 
 
-def array_namespace(*xs, api_version=None, use_compat=None) -> Namespace:
+def array_namespace(
+    *xs: Union[Array, bool, int, float, complex, None],
```
jorenham, line 441: it's sufficient to write

```suggestion
    *xs: Array | complex | None,
```

Keep in mind that in case of type-checker errors, often the entire signature is printed. So it's usually a good idea to avoid long annotations (in function signatures, but also in general).

array_api_compat/common/_linalg.py
```diff
@@ -15,50 +11,53 @@
...
+    Vh: Array
 
 # These functions are the same as their NumPy counterparts except they return
 # a namedtuple.
-def eigh(x: ndarray, /, xp, **kwargs) -> EighResult:
+def eigh(x: Array, /, xp: Namespace, **kwargs) -> EighResult:
     return EighResult(*xp.linalg.eigh(x, **kwargs))
 
-def qr(x: ndarray, /, xp, *, mode: Literal['reduced', 'complete'] = 'reduced',
+def qr(x: Array, /, xp: Namespace, *, mode: Literal['reduced', 'complete'] = 'reduced',
```
jorenham, line 45: eccentric formatting

array_api_compat/common/_linalg.py
```diff
@@ -69,12 +68,12 @@ def cholesky(x: ndarray, /, xp, *, upper: bool = False, **kwargs) -> ndarray:
 
 # The rtol keyword argument of matrix_rank() and pinv() is new from NumPy.
 # Note that it has a different semantic meaning from tol and rcond.
-def matrix_rank(x: ndarray,
+def matrix_rank(x: Array,
```
jorenham, line 71: eccentric formatting

array_api_compat/cupy/_aliases.py
```diff
@@ -69,20 +67,21 @@
...
-        ndarray,
+        Array,
         bool,
         int,
         float,
-        NestedSequence[bool | int | float],
+        complex,
+        NestedSequence[bool | int | float | complex],
         SupportsBufferProtocol,
     ],
```
jorenham, line 77: if I remember correctly, arrays are required to support the buffer protocol, so you could simplify this to:

```
    obj: float | NestedSequence[float] | SupportsBufferProtocol,
```

array_api_compat/cupy/_aliases.py
```diff
@@ -69,20 +67,21 @@
...
     /,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     copy: Optional[bool] = _copy_default,
```
jorenham, line 82: `_copy_default` is not assignable to `bool | None`. Since it's defined as an `object()`, the only way to correctly annotate `copy` is as:

```
    copy: object = _copy_default,
```

I'd suggest using a `enum.Enum` based sentinel here (originally proposed in https://github.com/python/typing/issues/236#issuecomment-227180301), which can properly be annotated:

```py
import enum
from typing import Final, Literal

class _Default(enum.Enum):
    copy = enum.auto()

_CopyDefault: TypeAlias = Literal[_Default.copy]
_copy_default: Final[_CopyDefault] = _Default.copy
```

which can then be used as

```
    copy: bool | _CopyDefault = _copy_default,
```

array_api_compat/dask/array/fft.py
```diff
@@ -16,9 +17,5 @@
 fftfreq = get_xp(da)(_fft.fftfreq)
 rfftfreq = get_xp(da)(_fft.rfftfreq)
 
-__all__ = [elem for elem in fft_all if elem != "annotations"] + ["fftfreq", "rfftfreq"]
-
-del get_xp
-del da
-del fft_all
-del _fft
+__all__ = fft_all + ["fftfreq", "rfftfreq"]
```
jorenham, line 20: static type-checkers don't support this

array_api_compat/numpy/_aliases.py
```diff
@@ -77,20 +74,21 @@ def _supports_buffer_protocol(obj):
...
+        complex,
+        NestedSequence[bool | int | float | complex],
         SupportsBufferProtocol,
     ],
     /,
     *,
-    dtype: Optional[Dtype] = None,
+    dtype: Optional[DType] = None,
     device: Optional[Device] = None,
     copy: "Optional[Union[bool, np._CopyMode]]" = None,
```
jorenham, line 89: no need for quotes here; `np._CopyMode` even exists at runtime.

array_api_compat/numpy/_aliases.py
```diff
@@ -77,20 +74,21 @@ def _supports_buffer_protocol(obj):
...
-        ndarray,
+        Array,
         bool,
         int,
         float,
-        NestedSequence[bool | int | float],
+        complex,
+        NestedSequence[bool | int | float | complex],
         SupportsBufferProtocol,
     ],
```
jorenham, line 84:
```
    obj: complex | NestedSequence[complex] | SupportsBufferProtocol,
```

assuming that arrays are required to implement the buffer protocol

array_api_compat/numpy/_typing.py
```diff
@@ -1,46 +1,33 @@
...
+            np.float32,
+            np.float64,
+            np.complex64,
+            np.complex128,
+            np.bool_,
+        ]
+    ]
+except TypeError:
+    # NumPy 1.x on Python 3.9 and 3.10
+    DType = np.dtype
```
jorenham, line 33: type-checker wouldn't know what to make of a `try/catch` like this:

```suggestion
if TYPE_CHECKING:
    DType: TypeAlias = np.dtype[
        np.bool_ 
        | np.integer 
        | np.single  # same as float32
        | np.double  # slightly broader than float64 in numpy==2.2.*
        | np.csingle  # same as complex64
        | np.cdouble  # slightly broader than complex128 in numpy==2.2.*
    ]
else:
    DType = np.dtype
```

and don't forget the `from typing import TYPE_CHECKING`

tests/test_all.py
```diff
@@ -15,6 +15,16 @@
 from ._helpers import import_, wrapped_libraries
 
 import pytest
+import typing
+
+TYPING_NAMES = {
```
jorenham, line 20: a `frozenset` might be a bit cleaner

setup.py
```diff
@@ -33,5 +33,8 @@
         "Programming Language :: Python :: 3.13",
         "License :: OSI Approved :: MIT License",
         "Operating System :: OS Independent",
```
jorenham, line 35: you could also add the `Typing :: Typed` PyPI classifier now

jorenham (commented):
Yay, typing! 🎉

Apart from the inline comments, there's also some generally applicable opportunities for improvement

- Using `TypeAlias` when defining things like `DType: TypeAlias = np.dtype[...]`, so that type-checkers know how to treat it
- Avoid dynamic `__all__` construction, so that static type-checkers know what's exported
- Since Python 3.9 `typing.Tuple` is deprecated in favor of using `tuple` directly
- `Optional` and `Union` are not needed when annotating things if you use `__future__.annotations`, even if you use Python 3.9.
- all `*args` and `**kwargs` I've seen are missing annotations. If unsure what these should be, then you should use `: object` for both (`Any` is not needed in input positions like these).
- `bool | int` simplifies to `int`, `int | float` simplifies to `float`, and `float | complex` simplifies to `complex`. Shorter annotations helps to avoid type-checker error message spaghetti.

array_api_compat/common/_typing.py
```diff
@@ -1,26 +1,24 @@
...
-
 _T_co = TypeVar("_T_co", covariant=True)
 
 class NestedSequence(Protocol[_T_co]):
     def __getitem__(self, key: int, /) -> _T_co | NestedSequence[_T_co]: ...
     def __len__(self, /) -> int: ...
 
-SupportsBufferProtocol = Any
 
+SupportsBufferProtocol = Any
```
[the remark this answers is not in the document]
NeilGirdhar: I thought this was a placeholder?  Aren't they ultimately going to point `SupportsBufferProtocol` to `array-api-typing`?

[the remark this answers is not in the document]
jorenham: > I thought this was a placeholder? Aren't they ultimately going to point `SupportsBufferProtocol` to `array-api-typing`?

My point is that there's no need to wait, because it can already be done without much effort

setup.py
NeilGirdhar: FYI I added it in https://github.com/data-apis/array-api-compat/pull/255

ev-br:
A naive question. Ideally, we'd use the same annotations for array api stubs, array-api-compat wrappers and array-api-strict objects. Can we actually have this kind of reuse? If yes in principle, what are the blockers and what's missing?
Then, it probably means the typing stubs live somewhere not inline---similar to scipy-stubs?

I suppose it doesn't matter all that much whether it matures here or in array-api-typing or somewhere else, as long as it does mature and gets reused. Wherever is more convenient, as long as we have some sort of a rough plan for the grand unification in the end.

array_api_compat/numpy/_typing.py
ev-br: please no `np.single`, `np.double` etc. Whenever these are used, the usage should drift towards explicit bit-width names, not the other way around.

jorenham:
> A naive question. Ideally, we'd use the same annotations for array api stubs, array-api-compat wrappers and array-api-strict objects. Can we actually have this kind of reuse? If yes in principle, what are the blockers and what's missing? Then, it probably means the typing stubs live somewhere not inline---similar to scipy-stubs?
>
> I suppose it doesn't matter all that much whether it matures here or in array-api-typing or somewhere else, as long as it does mature and gets reused. Wherever is more convenient, as long as we have some sort of a rough plan for the grand unification in the end.

Stubs target a specific package. What your talking about sounds more similar to [optype](https://github.com/jorenham/optype), which is a collection of re-usable protocols, aliases, and other typing utilities, that's heavy used by `scipy-stubs`. 

It's indeed that role that `array-api-typing` is intended to fulfill; a collection of types and typing-related utilities to help annotate array-api-related code. 

At the moment I'm busy rewriting the numpy stubs, and building an accompanying `optype`-like (optional) typing package for NumPy. Once that's off the ground, then I plan to redirect my focus towards array-api-typing. 

In the meantime, I don't see any need for holding off on writing `Protocols` yourself, because such a protocol might be provided by array-api-typing in the future. In fact, it will probably help speed up the development of array-api-typing that way, because that would give us a better idea of what works, and what doesn't.

NeilGirdhar:
Awesome, so ultimately most of these typing definitions will be lifted into array-api-typing, and this repo will point to it and depend on it?

Sounds like a great plan to me.  (Sorry if I'm so excited about this.  Can't wait to have my Array API code type checked!)

jorenham:
> Awesome, so ultimately most of these typing definitions will be lifted into array-api-typing, and this repo will point to it and depend on it? (And therefore, most of this PR will be deleted from this repo in the long run.)
> 
> Sounds like a great plan to me. (Sorry if I'm so excited about this. Can't wait to have my Array API code be type checked!)

Array-api-compat is still going to need annotations. But array-api-typing will make it easier to do this in a more accurate way.

crusaderky:
@jorenham, I really do appreciate your extensive review, but it goes _much_ below the intended scope of my PR. Most of your comments look like great material for a follow-up.
There are places where I added deliberately obsolete notation that is consist with the notation style that is extensively used across the repo, specifically because I didn't want to update the whole repo in a single PR.

I must confess that I'm struggling to pick out which of your comments request changes within the original scope of the PR, e.g. where you believe that _what the PR changes_ needs tweaking.

jorenham:
> @jorenham, I really do appreciate your extensive review, but it goes _much_ beyond the intended scope of my PR. Most of your comments look like great material for a follow-up. There are places where I added deliberately obsolete notation that is consistent with the notation style that is extensively used across the repo, specifically because I didn't want to update the whole repo in a single PR.
> 
> I must confess that I'm struggling to pick out which of your comments request changes within the original scope of the PR, e.g. where you believe that _what the PR changes_ needs tweaking.

Yea of course; most of my suggestions fall into the category of dotting the "i"'s. Specifically the stuff about overloads, the `**kwargs: object` and simplifications for `float | int` and the `NestedSequence` unions, are perfectly fine to ignore, and perhaps reconsider at a later point.

array_api_compat/cupy/_aliases.py
crusaderky: Do you mean `obj: complex | NestedSequence[complex] | SupportsBufferProtocol`?
I do worry about

jorenham: oh woops, yea, good catch

crusaderky:
> bool | int simplifies to int, int | float simplifies to float, and float | complex simplifies to complex. Shorter annotations helps to avoid type-checker error message spaghetti.

I can't find any mention that bool | int simplifies to int?
https://typing.readthedocs.io/en/latest/spec/special-types.html#special-cases-for-float-and-complex

array_api_compat/numpy/_aliases.py
crusaderky: They are not, because it's CPU-only.
It is explained here: https://data-apis.org/array-api/latest/design_topics/data_interchange.html

NeilGirdhar:
> I can't find any mention that bool | int simplifies to int?

Is [this](https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018/1) a reference?

crusaderky:
> > I can't find any mention that bool | int simplifies to int?
> 
> `bool` is already a subtype of `int` (which is unfortunate, in my opinion), so `bool | int` simplifies by subsumption.
> 
> For the others, is [this](https://discuss.python.org/t/clarifying-the-float-int-complex-special-case/54018/1) a reference?

That thread, while interesting, makes no mention of bool?

NeilGirdhar:
Sorry, maybe you missed my edit.  `bool` is already a subtype of `int` in Python, so `int | bool` is `int` because `int` subsumes `bool`.

jorenham:
> > bool | int simplifies to int, int | float simplifies to float, and float | complex simplifies to complex. Shorter annotations helps to avoid type-checker error message spaghetti.
> 
> I can't find any mention that bool | int simplifies to int? [typing.readthedocs.io/en/latest/spec/special-types.html#special-cases-for-float-and-complex](https://typing.readthedocs.io/en/latest/spec/special-types.html#special-cases-for-float-and-complex)

TLDR;

in case of `bool <: int`, it reflects the runtime behaviour:

```pycon
>>> issubclass(bool, int)
True
```

but in case of `int <: float` and `float <: complex`, it doesn't:

```pycon
>>> issubclass(int, float)
False
>>> issubclass(float, complex)
False
```

array_api_compat/common/_aliases.py
```diff
@@ -150,23 +161,23 @@ def zeros_like(
...
 # but I don't see any issue with exporting the names here regardless.
 class UniqueAllResult(NamedTuple):
-    values: ndarray
-    indices: ndarray
-    inverse_indices: ndarray
-    counts: ndarray
+    values: Array
+    indices: Array
+    inverse_indices: Array
+    counts: Array
```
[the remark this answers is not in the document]
crusaderky: Not something we get to decide on without completely breaking backwards compatibility with numpy

array_api_compat/numpy/_typing.py
jorenham: In most cases I agree. But here it's a sneaky workaround for the fact that on `numpy==2.2`, the `float64` is annotated as a proper subtype of `floating[_64Bit]`, while `double` is (still) an alias to `floating[_64Bit]`. 
The consequence is that `x: np.float64 = np.double()` will be rejected, since `double` is a broader type than `float64`. So if you want to accept *both* `double` and `float64`, then you should use `x: double`. Because if you use `x: float64`, then you're excluding `double` and anything else that resolves to `floating[_64Bit]`.

array_api_compat/numpy/_aliases.py
jorenham: Ok, good to know 👌🏻

jorenham:
> @jorenham I believe I addressed all comments _within the scope of this PR_, e.g. remarks on what this PR changes. Did I miss anything?

I believe that about covers it 👌🏻.

Do you want me to work on a follow-up that includes my remaining suggestions, once this is merged?

crusaderky:
> Do you want me to work on a follow-up that includes my remaining suggestions, once this is merged?

That would be great!

lucascolley (approved):
thanks @crusaderky, @jorenham ! Please feel free to open issues for follow-ups if that would be useful to track ideas.

lucascolley:
there is a merge conflict

ev-br:
> I can't find any mention that bool | int simplifies to int?

Please don't remove explicit bools. Python bools notwithstanding, Array API promotions (https://data-apis.org/array-api/2023.12/API_specification/type_promotion.html) do distinguish between `bool` and `int`.
So unless and until that changes in the spec, `array-api-compat` must treat them separately.

One other thing I'd really like to see is the same notation (DType vs Dtype, Array or array or ndarray) here and in `array-api-strict`. It's great that you made it uniform here @crusaderky, could we please have `-strict` follow.

crusaderky:
solved conflicts.
numpy 1.21 failure is unrelated to this PR.
ruff failure is an upstream regression: https://github.com/astral-sh/ruff/pull/16878

jorenham:
> Please don't remove explicit bools.

I agree that explicit is better than implicit. But keep in mind that unions can have a big impact on type-checker performance when used to annotate  parameters of overloaded functions. It also makes the error messages that type-checkers report longer, which makes typing errors more difficult to debug.

jorenham:
One last thing about `bool | int`: It can lead to unexpected typing errors, depending on which type-checker you use:

```py
import time
from typing import assert_type

def f[T1, T2](a: T1, b: T2) -> T1 | T2:
    return a if time.time() % 2 else b

b: bool = bool("")
i: int = int("42")

assert_type(f(b, i), int)
```

Here, mypy reports no errors, but `(based)?pyright` reports

```
"assert_type" mismatch: expected "int" but received "bool | int"  (reportAssertTypeFailure)
```

Playground links:

- [mypy](https://mypy-play.net/?mypy=latest&python=3.12&gist=53a97e1d53821bb24856c19b02e50821)
- [pyright](https://pyright-play.net/?strict=true&code=JYWwDg9gTgLgBDUBTAUAMyhECCeZgB2A5nKJLHAIYDO1SsA%2BjHqigCZJpxoDaAKgEYANHD4AmALoAKSgC5RwuACN54gJRwAtAD4FcAD6ixslHDNwoSGAFcoBKqS6IQSAHTOkUjQFI4YuEgANnTKKCgqyhAQgXAAvJHRUgBESWoowPKE8PFZyQAsYqlhNHSMzGCeaFJKIsBqtQQwaUA)
- [basedpyright](https://basedpyright.com/?typeCheckingMode=all&reportUnusedParameter=true&reportUnusedCallResult=false&code=JYWwDg9gTgLgBDUBTAsAKAGZQiBBPMYAOwHM5RJY4BDAZ1qVgH0YDU10ATJDODAbQAqARgA0cQQCYAugApqALgli4AIyVSAlHAC0APmVwAPhMkL0cS3ChIYAVyhEa5XohBIAdG6SztAUjhJOCQAGwY1dHR1NQgIELgAXhi42QAiVM10YCVieCTctIAWSQzItDoGZlYwHwxZVXFgTUaiGEy0IA)

jorenham:
> > Here, mypy reports no errors, but `(based)?pyright` reports
> > 
> > ```
> > "assert_type" mismatch: expected "int" but received "bool | int"  (reportAssertTypeFailure)
> > ```
> How is this not a bug?

In pyright there are no bugs; everything is "as designed" 🤭

> I'm +1 in favour of compactness myself. FWIW, `typing` is currently very prominently lacking "and" and  "not" operators, so it's impossible to write `complex & ~bool` or `complex and not bool`.

Amen. Intersection types is probably the no. 1 or no. 2 on my python typing wishlist. 

> Eventually, I'd hope for `array-api-types` to define Array as a generic that accepts bespoke types that mimic the isdtype argument, e.g. `Array[Integral | Floating]` (where `Integral` is a superclass of `SignedInteger` and `UnsignedInteger`, and  `Floating` is a superclass of `RealFloating` and `ComplexFloating`).
> How doubt that will integrate well with it being a `Protocol` though.

That's pretty much what I had in mind as well. And protocols are actually essential to building that, so no worries

ev-br:
> I agree that explicit is better than implicit. But keep in mind that unions can have a big impact on type-checker performance when used to annotate parameters of overloaded functions. It also makes the error messages that type-checkers report longer, which makes typing errors more difficult to debug.

>If you're new to static typing in python, and you see bool | int, it wouldn't be strange to think that annotating x: int would disallow bool (which is wrong).

Well, implementation details aside, the semantics of bools and ints is clearly different, at least in numeric contexts.
If the typing community does not recognize it, maybe it takes somebody immersed in that community (hint hint) to advocate for fixing this. Array API promotion rules, https://data-apis.org/array-api/2023.12/API_specification/type_promotion.html, can be thought of a distillation of a couple of decades of collective experience.

jorenham:
> Well, implementation details aside, the semantics of bools and ints is clearly different, at least in numeric contexts.
> If the typing community does not recognize it, maybe it takes somebody immersed in that community (hint hint) to advocate for fixing this.

I agree. In fact, I agree so much that I created a workaround for this: A type that only accepts `int` but rejects `bool`. And coincidentally, it's already part of `optype`, and is called `JustInt`. And as a cherry on top, it even has [documentation](https://github.com/jorenham/optype/blob/master/README.md#just)).

crusaderky:
I think this PR is ready to be merged?

lucascolley:
Ruff failure looks real?

jorenham:
> Ruff failure looks real?

It's looks similar to a bug in ruff 0.11.1, which has been fixed in 0.11.2

lucascolley:
thanks again Guido

array_api_compat/torch/_aliases.py
```diff
@@ -618,22 +611,22 @@ def linspace(start: Union[int, float],
...
+             **kwargs) -> Array:
     if not endpoint:
         return torch.linspace(start, stop, num+1, dtype=dtype, device=device, **kwargs)[:-1]
     return torch.linspace(start, stop, num, dtype=dtype, device=device, **kwargs)
 
 # torch.full does not accept an int size
 # https://github.com/pytorch/pytorch/issues/70906
 def full(shape: Union[int, Tuple[int, ...]],
-         fill_value: Union[bool, int, float, complex],
+         fill_value: complex,
```
ev-br, line 625: This and similar changes are still incorrect. Please revert this and other instances of this @lucascolley @crusaderky

lucascolley: It is a matter of preference as they are equivalent. I absolutely agree that we should be as explicit as possible in docstrings, but I am yet to be convinced that the benefits of being explicit in type annotations outweigh the drawbacks.

jorenham:
I'll be working on the follow-up

lucascolley:
To explain further my motivation for merge here: I think the problems highlighted with implicit subtypes are a (serious) issue for everyone unfamiliar with the nuances of Python typing, but they are problems that need to be addressed at a wider level. I think it would be a big mistake to remove this explicitness from the spec, as that is the source of truth for these APIs to which people will look. On the other hand, those looking at type annotations for array-api-compat are likely to be those more familiar with these typing nuances.

What we really need to make as clear as possible are the spec pages. I'm hoping that array-api-typing can provide concise, documented type aliases so that we can get rid of types like `Union[bool, int, float, complex]` from https://data-apis.org/array-api/draft/API_specification/generated/array_api.full.html. Aliases like `Scalar` and `RealScalar` and `NumericScalar` (which would be an alias for a union of `JustInt`, `JustFloat` and `JustComplex`) will make the spec read much better in my opinion.

Until we get there, I don't see the harm in simplifying signatures in these auxiliary libraries, especially when changes are (however begrudgingly) consistent with the wider typing ecosystem.

ev-br:
> but they are problems that need to be addressed at a wider level.

This alone is a smoking-gun sign that things should not be merged as is until there's a consensus, or at least a wider level attempt at reaching one.

lucascolley:
feel free to revert the merge if you would like to continue the discussion!

jorenham:
> e if you would like to continue the discussion!

I'd rather that you don't actually, as I'm quite far into the follow-up PR now

jorenham:
If we're going to conform to the Python typing spec, then `int | bool` simply isn't an option: https://typing.python.org/en/latest/reference/best_practices.html#types

If we don't, then we're need to have an exceptionally good reason for doing so. But so far, I haven't heard any.

crusaderky:
Can we open a high-level discussion on `array-api/` and reach a final consensus there with the wider community, instead of arguing among the 4 of us on individual PRs?

PR #261, opened 2025-02-27
BUG: `clip()` should not have `out=` parameter

crusaderky:
Fix issue where clip() would accept an out= parameter in numpy (wrapped or not), cupy(wrapped or not), and wrapped torch and in no other case.

I don't think that a deprecation cycle is advisable in this case.
I checked scipy and there are no use cases for this.

### DIFF
--- a/array_api_compat/common/_aliases.py
+++ b/array_api_compat/common/_aliases.py
@@ -331,8 +331,6 @@ def clip(
     max: Optional[Union[int, float, ndarray]] = None,
     *,
     xp,
-    # TODO: np.clip has other ufunc kwargs
-    out: Optional[ndarray] = None,
 ) -> ndarray:
     def _isscalar(a):
         return isinstance(a, (int, float, type(None)))
@@ -368,9 +366,9 @@ def clip(
     if type(max) is int and max >= wrapped_xp.iinfo(x.dtype).max:
         max = None
 
-    if out is None:
-        out = wrapped_xp.asarray(xp.broadcast_to(x, result_shape),
-                                 copy=True, device=device(x))
+    out = wrapped_xp.asarray(xp.broadcast_to(x, result_shape),
+                             copy=True, device=device(x))
+
     if min is not None:
         if is_torch_array(x) and x.dtype == xp.float64 and _isscalar(min):
             # Avoid loss of precision due to torch defaulting to float32


ev-br:
I don't think this is correct. `array-api-compat` generally allows additional arguments or other additional behaviors beyond what the spec allows. Restricting the behavior is what `array-api-strict` is for.

And this is by design: see "Avoid Restricting Behavior that is Outside the Scope of the Standard" section in 
https://data-apis.org/array-api-compat/dev/special-considerations.html

crusaderky:
Ah, ok.
`out=` was untested and broken.
I've fixed it and added a unit test.

array_api_compat/common/_aliases.py
```diff
@@ -368,23 +368,24 @@ def _isscalar(a):
     if type(max) is int and max >= wrapped_xp.iinfo(x.dtype).max:
         max = None
 
+    dev = device(x)
     if out is None:
-        out = wrapped_xp.asarray(xp.broadcast_to(x, result_shape),
-                                 copy=True, device=device(x))
+        out = wrapped_xp.empty(result_shape, dtype=x.dtype, device=dev)
+    out[()] = x
```
crusaderky, line 374: When out is None I expect this to have exactly the same performance as before

array_api_compat/common/_aliases.py
```diff
@@ -368,23 +368,24 @@ def _isscalar(a):
...
     if min is not None:
-        if is_torch_array(x) and x.dtype == xp.float64 and _isscalar(min):
-            # Avoid loss of precision due to torch defaulting to float32
-            min = wrapped_xp.asarray(min, dtype=xp.float64)
-        a = xp.broadcast_to(wrapped_xp.asarray(min, device=device(x)), result_shape)
+        a = wrapped_xp.asarray(min, dtype=x.dtype, device=dev)
+        a = xp.broadcast_to(a, result_shape)
         ia = (out < a) | xp.isnan(a)
         # torch requires an explicit cast here
-        out[ia] = wrapped_xp.astype(a[ia], out.dtype)
```
crusaderky, line 381: Removed unnecessary deep copy

array_api_compat/common/_aliases.py
```diff
@@ -368,23 +368,24 @@ def _isscalar(a):
...
+
     if min is not None:
-        if is_torch_array(x) and x.dtype == xp.float64 and _isscalar(min):
-            # Avoid loss of precision due to torch defaulting to float32
-            min = wrapped_xp.asarray(min, dtype=xp.float64)
-        a = xp.broadcast_to(wrapped_xp.asarray(min, device=device(x)), result_shape)
+        a = wrapped_xp.asarray(min, dtype=x.dtype, device=dev)
+        a = xp.broadcast_to(a, result_shape)
         ia = (out < a) | xp.isnan(a)
         # torch requires an explicit cast here
```
crusaderky: Could not reproduce

array_api_compat/common/_aliases.py
```diff
@@ -368,23 +368,23 @@ def _isscalar(a):
...
+
     if min is not None:
-        if is_torch_array(x) and x.dtype == xp.float64 and _isscalar(min):
-            # Avoid loss of precision due to torch defaulting to float32
-            min = wrapped_xp.asarray(min, dtype=xp.float64)
-        a = xp.broadcast_to(wrapped_xp.asarray(min, device=device(x)), result_shape)
+        a = wrapped_xp.asarray(min, dtype=x.dtype, device=dev)
+        a = xp.broadcast_to(a, result_shape)
         ia = (out < a) | xp.isnan(a)
-        # torch requires an explicit cast here
```
crusaderky, line 380: Could not reproduce

array_api_compat/common/_aliases.py
```diff
@@ -368,23 +368,23 @@ def _isscalar(a):
...
-        if is_torch_array(x) and x.dtype == xp.float64 and _isscalar(min):
-            # Avoid loss of precision due to torch defaulting to float32
-            min = wrapped_xp.asarray(min, dtype=xp.float64)
-        a = xp.broadcast_to(wrapped_xp.asarray(min, device=device(x)), result_shape)
-        ia = (out < a) | xp.isnan(a)
-        # torch requires an explicit cast here
-        out[ia] = wrapped_xp.astype(a[ia], out.dtype)
+        a = wrapped_xp.asarray(min, dtype=x.dtype, device=dev)
+        a = xp.broadcast_to(a, result_shape)
+        ia = ~(out >= a)  # Propagate NaNs
```
crusaderky, line 379: 2x faster than before, both for ints and floats, and now only 15% slower than just `out < a` for ints

ev-br:
Okay, the added test looks convincing, the CI gives a green light, the patch looks like a nice fix and a simplification. Let's roll with it. Thanks @crusaderky

PR #262, opened 2025-02-28
TST: run tests without all libraries installed

crusaderky:
Crucially, without this it's impossible to run on numpy 2.2, as sparse pins it to <=2.0


ev-br (approved):

ev-br:
Thanks @crusaderky , it is indeed annoying otherwise (or worse).

PR #263, opened 2025-02-28
TST: dask tests are extremely flaky

crusaderky:
Dask is running on 5 cases per test, causing it to be extremely flaky.
~~This PR brings it back in line with other libraries to 200.~~
This PR tweaks xfails and skips after running locally on 2000 examples and adds some commentary.

### DIFF
--- a/.github/workflows/array-api-tests-dask.yml
+++ b/.github/workflows/array-api-tests-dask.yml
@@ -10,4 +10,7 @@ jobs:
       package-version: '>= 2024.9.0'
       module-name: dask.array
       extra-requires: numpy
-      pytest-extra-args: --disable-deadline --max-examples=5
+      # Dask is quite slower then other libraries on unit tests
+      # Reduce the number of examples to speed up CI, even though
+      # this can cause flakiness.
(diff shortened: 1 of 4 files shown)


ev-br:
All other test runs are under five minutes, and even with `max-examples 5`, dask tests are the slowest. I'd rather strongly prefer all tests here to run under a few minutes.

Even the default 200 examples is more of a smoke-testing really. When looking for regressions or working on a new feature, I regularly run with `--max_examples=10_000`  locally. Not on dask, of course, that would never complete.

----------------------------

The state of the CI currently is not great indeed. One possibility is to have a nightly job which runs many examples.
One other idea is to export arrays in "unvectorized" tests to dlpack or something: https://github.com/data-apis/array-api-tests/issues/197#issuecomment-1806494046.

crusaderky:
This is very frustrating. I've reverted the number of examples back to 5 and ran locally with 2000 examples. The updated xfails and skips files reflect that.

ev-br:
Thanks @crusaderky !
It is frustrating, yes. Suggestions for improvements are most welcome indeed

Issue #230, opened 2025-01-07
Drop Python 3.9


crusaderky:
The latest version of several of the libraries that integrate with array-api-compat doesn't support Python 3.9 anymore. This produces the headache of having a de-facto, undocumented minimum supported version of those libraries, which is going to become more and more obsolete as time goes by, unwittingly enforced by the Python 3.9 CI jobs.

I think array-api-compat should follow the rest of the industry and formally drop python 3.9 support.

NeilGirdhar:
Does this project not follow [SPEC 0](https://scientific-python.org/specs/spec-0000/)?

ev-br:
Not automatically, no. 
Downstream consumers may or may not follow SPEC 0 for various reasons, and we should not exert an undue pressure on them. 

We discussed this in the last consortium meeting. The outcome is that we'll likely be dropping python 3.9 and numpy 1.21 after the next release.

PR #266, opened 2025-03-03
CI: drop python 3.9, numpy 1.21

ev-br:
closes gh-230

This likely affects scikit-learn, https://github.com/scikit-learn/scikit-learn/pull/30895#issuecomment-2690246905


lucascolley:
scikit-learn have now dropped 1.21

.github/workflows/array-api-tests.yml
```diff
@@ -40,9 +40,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        # Min version of dask we need dropped support for Python 3.9
-        # There is no numpy git tip for Python 3.9 or 3.10
-        python-version: ${{ (inputs.package-name == 'dask' && fromJson('[''3.10'', ''3.11'', ''3.12'']')) || (inputs.package-name == 'numpy' && inputs.xfails-file-extra == '-dev' && fromJson('[''3.11'', ''3.12'']')) || fromJson('[''3.9'', ''3.10'', ''3.11'', ''3.12'']') }}
+        python-version: ['3.11', '3.12', '3.13']
```
lucascolley, line 43: No 3.10?

lucascolley (commented):
Do we test with NuPy 1.22?

.github/workflows/array-api-tests.yml
ev-br: Yes, as an attempt to trim the CI size a bit. Experience over the last couple of months is that there were no issues which were python version dependent.

ev-br:
Am planning to merge this next week if this looks reasonable to people.  A possible contention point is the size of the CI matrix: this PR proposes to drop python 3.9 and 3.10 and numpy 1.21; this way we'll test on python 3.11, 3.12 and 3.13 x numpy 1.26, "latest released", and numpy-dev.

rgommers:
The linked scikit-learn issue has a clear plan, which is nice. It just dropped Python 3.9, so dropping Python 3.10 here seems quite aggressive - and probably not necessary? If you want to reduce the CI matrix, just remove 3.11 instead while keeping the lowest-supported version.

lucascolley:
Similarly, I think we should be testing NumPy 1.22 as the min. supported by sklearn.

lucascolley:
I spoke to Olivier and Tim yesterday and suggested that there shouldn't be any problems with array-api-compat and array-api-extra dropping more aggressively than scikit-learn, now that they have written down some reasonable rules. This matters to them as we are very close to them vendoring both libraries: https://github.com/scikit-learn/scikit-learn/pull/30340

lucascolley:
For comparison, in array-api-extra [we define the following test envs](https://github.com/data-apis/array-api-extra/blob/9631f3289f2bb2e261baa74283bef4c5adf2a270/pyproject.toml#L175-L180)

```toml
tests-numpy1 = ["py310", "tests", "numpy1"]
tests-py310 = ["py310", "tests"]
tests-py313 = ["py313", "tests"]
tests-backends = ["py310", "tests", "backends"]
tests-cuda = ["py310", "tests", "backends", "cuda-backends"]
```

and use the non-CUDA envs in [the CI matrix](https://github.com/data-apis/array-api-extra/blob/9631f3289f2bb2e261baa74283bef4c5adf2a270/.github/workflows/ci.yml#L50-L51)

```yml
matrix:
    environment: [tests-py310, tests-py313, tests-numpy1, tests-backends]
```

Our `numpy1` env is currently using oldest SPEC 0 NumPy (1.25), but I think we should change that to the oldest supported by sklearn (1.22) (https://github.com/data-apis/array-api-extra/pull/169#issuecomment-2743095328)

.github/workflows/array-api-tests.yml
ev-br, line 43:
```suggestion
        python-version: ['3.10', '3.12', '3.13']
```

.github/workflows/tests.yml
```diff
@@ -5,13 +5,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
-        numpy-version: ['1.21', '1.26', '2.0', 'dev']
+        python-version: ['3.11', '3.12']
```
ev-br, line 8:
```suggestion
        python-version: ['3.10', '3.12']
```

lucascolley (changes requested):
to clarify, I would like to see NumPy 1.22 tested before merging

ev-br:
Well, from past experience there was not much useful signal from keeping both numpy 1.21 and 1.26.  All it did was a bit was a bit more churn to duplicate xfails for the two numpy versions. If we are to add 1.22, we can keep 1.21 just as well.

lucascolley:
How about 1.22 instead of 1.26?

ev-br:
Honestly, I don't see much point. 
Esp given that we're planning to make parts of the testing matrix only compatible with python 3.12+ for `_buffer__`, 
https://hackmd.io/zn5bvdZTQIeJmb3RW1B-8g#Downstream-libraries-update, and numpy 1.26 seems to be the first one to even support python 3.12.
Granted, it's mainly for array-api-strict, not array-api-compat. So technically, we can keep -compat very very conservative. Why though.

lucascolley:
Well, I suppose we should ask the scikit-learn devs, as they are the 'users' of the NumPy 1.22 support. Does you foresee it being useful to catch things in array-api-compat CI before they propagate to array-api-extra or scikit-learn, @betatim.

crusaderky:
My 2 cents on NumPy 1.22 is that
1. it's out of [SPEC0](https://scientific-python.org/specs/spec-0000/) so it's reasonable to ask scikit-learn developers to drop it. I think it would also be reasonable to say that they won't be able to upgrade array-api-compat beyond 1.11 until they do so.
2. however, if it is supported, it must be tested.

FYI SPEC0 dropped NumPy 1.25 two days ago.

I just verified that NumPy 1.22 does have `_CopyMode` and would work without this special casing, whereas 1.21 doesn't:
https://github.com/data-apis/array-api-compat/blob/b6900dfafd5a840399301be266aa17cd7d12cce3/array_api_compat/numpy/_aliases.py#L101-L112
that said, I don't know (without running the unit tests) if anything else would still require a special code path.

.github/workflows/tests.yml
```diff
@@ -5,13 +5,11 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
-        numpy-version: ['1.21', '1.26', '2.0', 'dev']
+        python-version: ['3.10', '3.12']
```
crusaderky, line 8:
```suggestion
        python-version: ['3.10', '3.12', '3.13']
```

lucascolley: 3.12 testing can be dropped from CI?

crusaderky:
> https://scikit-learn.org/dev/developers/maintainer.html#guideline-for-bumping-minimum-versions-of-our-dependencies

Forget about 1.22 vs. 1.26. This policy implies that they commit to support NumPy 1.x until October 2027, when Python 3.12 will fall out of their 4 years window. By then, SPEC0 will recommend NumPy >=2.3.
This is _a lot_ of luggage that everyone else who maintains their dependencies is asked to carry too.

NEP29 first and SPEC0 then were written for a reason. I find it a bit baffling - not that one library can decide to diverge (to each the problems they choose for themselves) - but that _other_ libraries, whose developers likely disagree with their decision, are asked to accommodate them.

.github/workflows/tests.yml
crusaderky: not if you want to test numpy 1.26 on the most recent python version it supports.

betatim:
I'm not sure I can add much here (sorry for being a few days late Lucas). scikit-learn will follow its policy for the next while. Both because it has been its inofficial policy for a long time and because getting agreement on making it official policy took some effort. Of course in the future things will get reconsidered and we left an escape hatch to drop things earlier if there is a good reason. Right now I think it will be hard to convince the rest of scikit-learn to accelerate the Numpy schedule because of a compat library that is used by an experimental feature of scikit-learn (at least I'm hard to convince).

I think having a CI job that tests the oldest set of dependencies is worth having, and skipping intermediate versions is a good way to reduce CI matrix size.

Issue #274, opened 2025-03-16
`torch.result_type` does not accept combinations of scalars and dtypes


mdhaber:
`torch.result_type` does not accept combinations of scalars and dtypes. (These are permitted by the standard; only one of the arguments has to be an array or dtype.)

```python3
from array_api_compat import torch as xp

types = ['scalar', 'array', 'dtype']
dtypes = [float, int, complex]

for type_a in types:
    for type_b in types:
        for dtype_a in dtypes:
            for dtype_b in dtypes:
                scalar_a = dtype_a(1)
                scalar_b = dtype_b(1)
                array_a = xp.asarray(scalar_a)
                array_b = xp.asarray(scalar_b)
                in1 = dict(scalar=scalar_a, array=array_a, dtype=array_a.dtype)[type_a]
                in2 = dict(scalar=scalar_b, array=array_b, dtype=array_b.dtype)[type_b]
                try:
                    xp.result_type(in1, in2)
                except:
                    print(in1, in2)
```

Result:
```
1.0 torch.float32
1.0 torch.int64
1.0 torch.complex64
1 torch.float32
1 torch.int64
1 torch.complex64
(1+0j) torch.float32
(1+0j) torch.int64
(1+0j) torch.complex64
torch.float32 1.0
torch.float32 1
torch.float32 (1+0j)
torch.int64 1.0
torch.int64 1
torch.int64 (1+0j)
torch.complex64 1.0
torch.complex64 1
torch.complex64 (1+0j)
```
This seems to include any combination of Python scalar and torch dtype and *only* these combinations. (E.g., array/scalar and scalar/array are fine).

Issue #273, opened 2025-03-16
`torch.result_type` results are order-dependent when default dtype is `float64`


mdhaber:
`torch.result_type` results are order-dependent when default dtype is `float64`.

```python3
from array_api_compat import torch as xp
xp.set_default_dtype(xp.float64)

dtypes = ['float32', 'float64', 'int32', 'int64', 'complex64', 'complex128']
for dtype_a in dtypes:
    for dtype_b in dtypes:
        a = xp.asarray([2, 1], dtype=getattr(xp, dtype_a))
        b = xp.asarray([1, -1], dtype=getattr(xp, dtype_b))
        dtype_1 = xp.result_type(a, b, 1.0)
        dtype_2 = xp.result_type(b, a, 1.0)
        if dtype_1 != dtype_2:
            print('dtype_1 != dtype_2:', dtype_1, dtype_2)

# dtype_1 != dtype_2: torch.float64 torch.float32
# dtype_1 != dtype_2: torch.float64 torch.float32
# dtype_1 != dtype_2: torch.float32 torch.float64
# dtype_1 != dtype_2: torch.complex64 torch.complex128
# dtype_1 != dtype_2: torch.float32 torch.float64
# dtype_1 != dtype_2: torch.complex64 torch.complex128
# dtype_1 != dtype_2: torch.complex128 torch.complex64
# dtype_1 != dtype_2: torch.complex128 torch.complex64
```

Thanks @ev-br for figuring out that it was dependent on the default dtype!

ev-br:
The root cause seems to be this:

```
In [1]: import torch

In [2]: torch.set_default_dtype(torch.float64)

In [3]: dt_a, dt_b = torch.float32, torch.int32

In [4]: a, b = torch.empty(1, dtype=dt_a), torch.empty(1, dtype=dt_b)

In [5]: torch.result_type(a, 1.0)
Out[5]: torch.float32

In [6]: torch.result_type(b, 1.0)
Out[6]: torch.float64
```

This is, strictly speaking, unspecified by the Array API spec because it combines a python float and an integer array. 
This is still something we should fix in `array-api-compat`, and relatively quickly. 
I think we'll release `array-api-compat==1.11.2` once https://github.com/data-apis/array-api-tests/pull/349 and https://github.com/data-apis/array-api-compat/pull/277 land.

PR #277, opened 2025-03-17
BUG: torch: fix `result_type` with python scalars

ev-br:
fixes gh-273,fixes gh-274

cross-ref https://github.com/data-apis/array-api-tests/pull/349 which adds tests.

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from functools import wraps as _wraps
+from functools import reduce as _reduce, wraps as _wraps
 from builtins import all as _builtin_all, any as _builtin_any
 
 from ..common import _aliases
@@ -124,25 +124,43 @@ _py_scalars = (bool, int, float, complex)
 
 
 def result_type(*arrays_and_dtypes: Union[array, Dtype, bool, int, float, complex]) -> Dtype:
-    if len(arrays_and_dtypes) == 0:
-        raise TypeError("At least one array or dtype must be provided")
-    if len(arrays_and_dtypes) == 1:
+    num = len(arrays_and_dtypes)
+
+    if num == 0:
+        raise ValueError("At least one array or dtype must be provided")
+
+    elif num == 1:
         x = arrays_and_dtypes[0]
         if isinstance(x, torch.dtype):
             return x
         return x.dtype
-    if len(arrays_and_dtypes) > 2:
-        return result_type(arrays_and_dtypes[0], result_type(*arrays_and_dtypes[1:]))
 
-    x, y = arrays_and_dtypes
-    if isinstance(x, _py_scalars) or isinstance(y, _py_scalars):
-        return torch.result_type(x, y)
+    if num == 2:
+        x, y = arrays_and_dtypes
+        return _result_type(x, y)
 
-    xdt = x.dtype if not isinstance(x, torch.dtype) else x
-    ydt = y.dtype if not isinstance(y, torch.dtype) else y
+    else:
+        # sort scalars so that they are treated last
+        scalars, others = [], []
+        for x in arrays_and_dtypes:
+            if isinstance(x, _py_scalars):
+                scalars.append(x)
+            else:
+                others.append(x)
+        if not others:
+            raise ValueError("At least one array or dtype must be provided")
 
-    if (xdt, ydt) in _promotion_table:
-        return _promotion_table[xdt, ydt]
+        # combine left-to-right
+        return _reduce(_result_type, others + scalars)
+
+
+def _result_type(x, y):
+    if not (isinstance(x, _py_scalars) or isinstance(y, _py_scalars)):
+        xdt = x.dtype if not isinstance(x, torch.dtype) else x
+        ydt = y.dtype if not isinstance(y, torch.dtype) else y
+
+        if (xdt, ydt) in _promotion_table:
+            return _promotion_table[xdt, ydt]
 
     # This doesn't result_type(dtype, dtype) for non-array API dtypes
     # because torch.result_type only accepts tensors. This does however, allow
@@ -150,6 +168,7 @@ def result_type(*arrays_and_dtypes: Union[array, Dtype, bool, int, float, comple
     x = torch.tensor([], dtype=x) if isinstance(x, torch.dtype) else x
     y = torch.tensor([], dtype=y) if isinstance(y, torch.dtype) else y
     return torch.result_type(x, y)
+
 
 def can_cast(from_: Union[Dtype, array], to: Dtype, /) -> bool:
     if not isinstance(from_, torch.dtype):
--- a/tests/test_all.py
+++ b/tests/test_all.py
@@ -16,6 +16,7 @@ from ._helpers import import_, wrapped_libraries
 
 import pytest
 
+@pytest.mark.skip(reason="TODO: starts failing after adding test_torch.py in gh-277")
 @pytest.mark.parametrize("library", ["common"] + wrapped_libraries)
 def test_all(library):
     if library == "common":
--- a/tests/test_torch.py
+++ b/tests/test_torch.py
@@ -1,0 +1,98 @@
+"""Test "unspecified" behavior which we cannot easily test in the Array API test suite.
+"""
+import itertools
+
+import pytest
+import torch
+
+from array_api_compat import torch as xp
+
+
+class TestResultType:
+    def test_empty(self):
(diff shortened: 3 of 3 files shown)


ev-br:
cc @mdhaber

array_api_compat/torch/_aliases.py
```diff
@@ -135,8 +135,12 @@ def result_type(*arrays_and_dtypes: Union[array, Dtype, bool, int, float, comple
         return result_type(arrays_and_dtypes[0], result_type(*arrays_and_dtypes[1:]))
 
     x, y = arrays_and_dtypes
-    if isinstance(x, _py_scalars) or isinstance(y, _py_scalars):
-        return torch.result_type(x, y)
+    if isinstance(x, _py_scalars):
+        if isinstance(y, _py_scalars):
+            raise ValueError("At least one array or dtype is required.")
```
mdhaber, line 140: If I'm understanding the recursion correctly, won't this fail if, for instance, `xp.result_type(xp.float64, 1, 2)` because neither of the last two arguments is an array or dtype?

seberg: True, but you must fundamentally change the logic here anyway.  While torch avoids tricky promotions (unlike NumPy) and promotions are associative because of that, you are currently breaking that by adding scalar promotion rules.
That isn't hard to fix, because you can fix it by fixing the binary promotion to:
```
promote(pythong_float, torch_integer) -> python_float
```

I.e. the recursive part of the algorithm *must* drag around `python_float` as a proper dtype.  Only at the very end can you convert it to the default float.

ev-br: Given that one way or other, we need to add an arbitrary rule, and we cannot fully satisfy all of Array API promotion rules, pytorch promotion rules and associativity, how about always sorting the inputs to put all scalars to the left. This way, the pytorch addition only kicks up in a binary promotion ("a conforming library may support additional type promotion rules", so pytorch does---for binary promotions only), and we definitely have associativity. Thoughts?

seberg: I don't understand the first premise of needing an arbitrary rule or that you can't satisfy everything...

... but it doesn't matter?  Currently, torch (by its rules) _is_ associative, I believe.  The only odd one out is the custom behavior for `float` due to `int_dtype + float -> default_float` being wrong (breaking associativity).

And sure, since that is all you need to deal with, you can deal with it by ensuring Python scalar promotions happen last.

seberg (commented):
Sorry had some other comments, meant to make a single "review".

(The point about being associative is, was that for the binary operation `result_type(a, result_type(b, c)) == result_type(result_type(a, b), c)`, if you ignore python `float`.)

ev-br:
Okay, third time the charm, hopefully.

Now that I look closely, torch in fact has a `result_type(Number, Number)` overload. So there's no need for sorting etc, just some gymnastics to map from `*arrays_and_dtypes_or_scalars` to torch's `Tensor | Number` arguments.

> Maybe since that is "unspecified" and because of that completely untested?

Exactly.

Which is why previous iterations attempted to only add minimal changes on top of what already was effectively tested by usage. That did not work, so here's a (still minimal) rework + some light smoke-testing of "unspecified" logic.

array_api_compat/torch/_aliases.py
```diff
@@ -131,12 +131,30 @@ def result_type(*arrays_and_dtypes: Union[array, Dtype, bool, int, float, comple
...
+        for x in arrays_and_dtypes:
+            if isinstance(x, _py_scalars):
+                scalars.append(x)
+            else:
+                others.append(x)
+        if len(scalars) == len(arrays_and_dtypes):
+            raise ValueError("At least one array or dtype is required.")
+
+        arrays_and_dtypes = scalars + others
         return result_type(arrays_and_dtypes[0], result_type(*arrays_and_dtypes[1:]))
```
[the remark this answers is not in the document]
ev-br: Yeah, thanks. The last commit removes all sorting and ditches recursion in favor of `functools.reduce`.

seberg:
Looks like nice code improvements, but:
> Now that I look closely, torch in fact has a result_type(Number, Number) overload. So there's no need for sorting

is incorrect.  In making everything nicer, the current code doesn't actually fix the original issue :)?!
The point is that the `result_type(int_tensor, 1.0)` will return `float64` which *breaks* associativity of `_result_type`.

You still need the sorting if you want order to not matter and you should maybe add a test for that:
```
    result_type([1., 5, 3, float16(3), 5, 6, 1.]) == float16
```
(or stay within that `float64` default dtype context and use float32)

seberg:
Yeah, I agree, you can choose not to guarantee that order doesn't matter!  I am pretty sure that torch promotion is associative, so that it isn't order dependent (besides for Python `float`). 

I thought that was the original issue/goal here to make order not matter in this context :).

ev-br:
Okay, thanks for clarifying it! Let's try to follow NumPy here and sorting back.

ev-br:
I'm going to skip `tests/test_all.py::test_all` for now. The test starts failing on CI (but not locally) when I add an unrelated test module (`tests/test_torch.py`); the test itself does something rather strange with direct access to `sys.modules`, and loops over private implementation modules. I frankly cannot tell what the intention is. Why do we care if `common/_helpers.py` has matching `__all__` and `dir()`? Also it picks up `common/_alises.py` on CI but misses it locally. 

If the intention is to check that high-level modules have correct `__all__` lists, then that's what the test should proble, I suppose.

ev-br:
Test weirdness aside, I think the current version is as good as it gets. Does this look reasonable to you @seberg?
Would be great if you could test it @mdhaber .
Unless somebody sees further problems, I'll keep it open for a short while, then merge and proceed.

seberg:
Yes, LGTM.  I can always find more nits if you like though ;p.
(Behavior nit: `result_type(1, 1)` doesn't raise, but `result_type(1, 1, 1)` does.)

ev-br:
> (Behavior nit: result_type(1, 1) doesn't raise, but result_type(1, 1, 1) does.)

Yeah. We cannot prohibit the former because torch allows it; allowing the latter goes against "no default dtype" stance as I understand it --- heck, even `where(x > 0, 1, 2)` is not allowed with two scalars.
So I'd rather block it, at least until pytorch defines `result_type(1, 1, 1)` if it ever does.
This is actually what I meant in an "arbitrary rules"  in https://github.com/data-apis/array-api-compat/pull/277#discussion_r2000643322. 

> I can always find more nits if you like though ;p.

Nope, PR review graphs are directed and nits only flow along one direction :-).

mdhaber:
Can we just merge it? Then I can just change the commit of `array-api-compat` that SciPy is using and see whether it passes tests.

ev-br:
Okay, this is okay'd by Sebastian, and has been tested via https://github.com/data-apis/array-api-tests/pull/349 locally, so hopefully it just works.
Merging.
Thanks @seberg for the (multiple-stage) review!

mdhaber:
I think this did the trick! I am still having one problem that I thought was due to gh-273, but maybe not - I'll have to take a closer look. Thanks!

mdhaber:
The one remaining problem was a different issue, but I found another: if `torch.result_type` with two Python scalar arguments doesn't raise, and we get an unexpected type of error if there is only one Python scalar argument.

```python3
from array_api_compat import torch as xp

xp.result_type(1)
# AttributeError: 'int' object has no attribute 'dtype'

xp.result_type(1, 1)
# torch.int64

xp.result_type(1, 1, 1)
# ValueError: At least one array or dtype must be provided
```

While this behavior is not governed by the standard, it would be helpful for testing purposes if it were self-consistent.

ev-br:
Ah yes, this was noted by Sebastian: https://github.com/data-apis/array-api-compat/pull/277#issuecomment-2737387294
I stand by https://github.com/data-apis/array-api-compat/pull/277#issuecomment-2737461639 : we cannot make it truly consistent, sadly.

For a one-argument version, turning the AttributeError into a ValueError is totally possible of course; PR welcome if you feel strongly enough!

seberg:
I think you could check for `result_type(1, 1)` and raise an error?  If that is less confusing.

ev-br:
That would make a precedent for `array-api-compat` raising on operations which are _unspecified_ in the spec but allowed by the array provider.

And I do believe that the binary `result_type(x, y)` is way more common in practice than `result_type(x, y, z, ....)`, so from a pragmatic POV the status quo is a lesser evil.

mdhaber:
Maybe we need an `array-api-compat-compat` that does this sort of thing.

Issue #283, opened 2025-03-21
skipped test_all


ev-br:
https://github.com/data-apis/array-api-compat/pull/277/commits/5473d84d5c36b23e091b880279c863c32f41b828 skipped a test, `tests/test_all.py`, which started failing after adding an unrelated `tests/test_torch.py`.
The test itself seems to probe the contents of internal implementation modules, and fails essientially on this:

```
In [16]: from array_api_compat.common import _aliases

In [17]: all_names = set(n for n in _aliases.__all__ if not n.startswith('_'))

In [18]: dir_names = set(n for n in dir(_aliases) if not n.startswith('_'))

In [19]: dir_names - all_names      # not empty, failure
Out[19]: 
{'NamedTuple',
 'TYPE_CHECKING',
 'annotations',
 'array_namespace',
 'device',
 'inspect',
 'is_cupy_namespace'}
```

I've to admit I'm not sure what the intention is, this seems to be a Chesterton fence of sorts. Do you happen to remember the context @asmeurer ?

asmeurer:
Yes, that test ensure that the `__all__` variable is defined correctly. We are defining it in a way that makes it difficult to manage in the ways you usually would, since we re-export names from existing packages, so we can't just manually list out every name that should be included. I would suggest keeping this test working. There's more context here https://github.com/data-apis/array-api-compat/pull/95.

Note that if a name is in a module that shouldn't be exported you can add it to `_all_ignore`.

ev-br:
Thanks @asmeurer ! 

So `common._aliases.__all__` was "wrong" for a while, but testing did not pick it up until an unrelated test actually imported the module. gh-286 adds a fix to `__all__`, plus a couple of small tweaks to make it less likely to miss a problem going forward. Would appreciate if you could take a look-through @asmeurer

PR #286, opened 2025-03-22
Reenable test_all, fix `_aliases.__all__`

ev-br:
closes gh-283

- reenable `test_all`
- fix `common._aliases.__all__` which was wrong
- make it easier to run self-tests locally, esp if not all optional dependencies are available.

So the `__all__` list was wrong for a while, but it was not caught be tests because:
- locally, self-tests required all optional dependencies to be installed, so I never ran the whole `$ pytest tests/` locally
- on CI, the problem was not caught either, because the test iterates over `sys.modules`. So if an implementation module is not imported, it's not checked: this was the case with `common._aliases` until gh-277 added an unrelated test.


tests/test_torch.py
```diff
@@ -3,7 +3,11 @@
 import itertools
 
 import pytest
-import torch
+
+try:
+    import torch
+except ImportError:
+    pytestmark = pytest.skip(allow_module_level=True, reason="dask not found")
```
ev-br, line 10:
```suggestion
    pytestmark = pytest.skip(allow_module_level=True, reason="pytorch not found")
```

ev-br:
I think I might want to merge this PR sooner than later, if only to restore the state of the tests and make the CI green by default. Please speak up if you want some more time to review.

tests/test_all.py
```diff
@@ -26,15 +26,14 @@
...
-@pytest.mark.skip(reason="TODO: starts failing after adding test_torch.py in gh-277")
 @pytest.mark.parametrize("library", ["common"] + wrapped_libraries)
 def test_all(library):
     if library == "common":
         import array_api_compat.common  # noqa: F401
     else:
         import_(library, wrapper=True)
 
-    for mod_name in sys.modules:
+    for mod_name in sys.modules.copy():
```
lucascolley, line 36: for my understanding, why is this needed? Perhaps we can add a code comment?

lucascolley (approved):
thanks Evgeni! Just one comment

tests/test_all.py
ev-br: Something imports something in the loop, and it otherwise fails with a `RuntimeError: dictionary changed size during iteration`.

ev-br:
Okay, let's merge this. Thanks Aaron for weighing in in https://github.com/data-apis/array-api-compat/issues/283, thanks Lucas for the review.

PR #290, opened 2025-03-25
ENH: correct Dask capabilities

crusaderky:


crusaderky:
All CI failures are unrelated.
I've run array-api-tests with `--max-examples=200` and got all green.
@ev-br ready for review and merge.

array_api_compat/dask/array/_info.py
```diff
@@ -68,11 +68,22 @@ def capabilities(self):
         The resulting dictionary has the following keys:
 
         - **"boolean indexing"**: boolean indicating whether an array library
-          supports boolean indexing. Always ``False`` for Dask.
+          supports boolean indexing.
+
+          Dask support boolean indexing as long as both the index
+          and the indexed arrays have known shapes.
+          Note however that the output .shape and .size properties
+          will contain a non-compliant math.nan instead of None.
```
ev-br, line 76: Is this something `array_api_compat.dask` could/should cover?

crusaderky: That would be impossible, as it is an attribute of the Array object. array-api-compat (wisely) has an explicit policy against subclassing/incapsulating.

Changing it upstream would also require an extensive test review and a deprecation cycle of opt-in None->opt-out None->hardcoded None, as such a change will break a lot of things, internally, in downstream libraries, and in final users' code.

array_api_compat offers a `size()` function to deal with it. We could add a `shape()` one but so far there hasn't been a pressing need for one, neither in scipy nor in array-api-extra.

ev-br: Ah, indeed. Fixing an array method is a no-go indeed.

Adding an `xp.shape` function might be a nice way out indeed. Whether catering for a dask idiosyncrasy is a strong enough motivation for it, I don't know. 

Anyhow, there's no need to block this PR on this then.

ev-br (approved):
Merging. Thanks Guido.

PR #297, opened 2025-04-03
TST: bump to ndonnx 0.10.1

crusaderky:


ev-br:
While I don't understand what this does or why this is needed, if it helps unbreak the CI, fine. Thanks Guido.

crusaderky:
> While I don't understand what this does or why this is needed, if it helps unbreak the CI, fine. Thanks Guido.

It tests `asarray` calls between backends. Which are not part of the Array API standard, unless the source backend happens to support the buffer protocol.

PR #253, opened 2025-02-20
ENH: torch: unsigned types

ev-br:
- [x] add type promotion rules for (uintN, uintM) and (uintN, intM)
- [x] return correct dtype lists in `_info`
- [x] stop skipping uintN tests on CI

EDIT: this PR is not mergeable currently.  Running tests torch/uint results in ~70 failures as of pytorch 2.6.0, see the CI logs below (and/or the fold-out in https://github.com/data-apis/array-api-compat/pull/253#issuecomment-2681396923). We have to wait for pytorch itself to proceed with `array-api-compat.torch` wrappers.

PR #298, opened 2025-04-03
ENH: `torch` dtype promotions

crusaderky:
- On PyTorch >=2.3, support uint16, uint32, and uint64 in `result_type`
- On PyTorch >=2.3, `sum` and `prod` will now promote uint8, uint16 and uint32 to uint64.
- Change `sum(x, axis=(), dtype=x.dtype)` and `prod` with the same parameters, which were previously returning `x` itself, to return a deep copy instead. While this is not part of the Array API, it's what both numpy an base torch do, and probably the most healthy thing to do xref https://data-apis.org/array-api/latest/design_topics/copies_views_and_mutation.html


ev-br:
https://github.com/data-apis/array-api-compat/pull/253 seems to have run into a wall with uint promotions in pytorch, did something change meanwhile?

crusaderky:
> #253 seems to have run into a wall with uint promotions in pytorch, did something change meanwhile in the meantime?

No, nothing has changed on the torch side AFAIK.

- The changes to the dtypes table in this PR are functionally the same as in #253, but IMHO cleaner. They don't need to wait for full torch support in various functions to be merged.
- The fix to `sum(x, dtype=x.dtype, axis=()` is unrelated to dtype promotion
- The promotion of uint8/16/32 to uint64 is missing in #253, and something you want to have anyway
- Unlike #253, this PR changes neither CI nor `__array_namespace_info__`, which is what makes that PR non-mergeable.

ev-br:
> this PR changes neither CI nor __array_namespace_info__, which is what makes that PR non-mergeable.

You lost me here :-). So is it basically "let's add something because we want it, but not test it on CI because we know it'll break"?

crusaderky:
> > this PR changes neither CI nor **array_namespace_info**, which is what makes that PR non-mergeable.
> 
> You lost me here :-). So is it basically "let's add something because we want it, but not test it on CI because we know it'll break"?

Well, we will need this and the other PR eventually when torch introduces support.
We can either keep this PR unmerged, or leave it in main but untested by CI. Final users that somehow manage to avoid the many non-functioning APIs will benefit from it immediately.

ev-br:
> We can either keep this PR unmerged, or leave it in main but untested by CI. 

The latter we definitely do not want to do.

crusaderky:
I've updated this to incorporate the latest changes and added the contents of #253, which can now be closed.

PR #299, opened 2025-04-03
BUG: `torch.asarray` device propagation

crusaderky:
+ Workaround to https://github.com/pytorch/pytorch/issues/150199
+ Also fixes device propagation in `prod`, `sum`, `any` and `all`, which had the same issue induced by array-api-compat


lucascolley (approved):
LGTM, thanks!

PR #301, opened 2025-04-03
MAINT: validate device on numpy and dask

crusaderky:
Salvaged from #293.

- Validate `device` keyword in `array_api_compat.numpy.astype`
- Validate `device` keyword in all Dask functions


ev-br:
Thanks @crusaderky

PR #302, opened 2025-04-03
BUG: Don't import helpers in namespaces

crusaderky:
Remove helper functions from wrapped array namespaces.
This was introduced when the numpy namespace was split from the helpers in 2022 (b3a12d92e840be6ef0f87f568d68a4f58b55c7e7).
This PR prompted by a collision between `array_api_compat.device` and `torch.device`.

Re. deprecation layer: unsure if we need one? This was never advertised in the documentation.

### DIFF
--- a/array_api_compat/common/_linalg.py
+++ b/array_api_compat/common/_linalg.py
@@ -174,3 +174,5 @@ __all__ = ['cross', 'matmul', 'outer', 'tensordot', 'EighResult',
            'svd', 'cholesky', 'matrix_rank', 'pinv', 'matrix_norm',
            'matrix_transpose', 'svdvals', 'vecdot', 'vector_norm', 'diagonal',
            'trace']
+
+_all_ignore = ['math', 'normalize_axis_tuple', 'get_xp', 'np', 'isdtype']
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -276,7 +276,7 @@ def test_asarray_copy(library):
     is_lib_func = globals()[is_array_functions[library]]
     all = xp.all if library != 'dask.array' else lambda x: xp.all(x).compute()
 
-    if library == 'numpy' and xp.__version__[0] < '2' and not hasattr(xp, '_CopyMode') :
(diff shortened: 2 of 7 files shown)


array_api_compat/common/_linalg.py
```diff
@@ -174,3 +174,5 @@ def trace(
            'svd', 'cholesky', 'matrix_rank', 'pinv', 'matrix_norm',
            'matrix_transpose', 'svdvals', 'vecdot', 'vector_norm', 'diagonal',
            'trace']
+
+_all_ignore = ['math', 'normalize_axis_tuple', 'get_xp', 'np', 'isdtype']
```
crusaderky, line 178: No idea why `test_all` wasn't failing before.

crusaderky:
This is ready for review.
I'm personally against building a cumbersome deprecation layer.
Nominally, because I expect that anybody who tests vs. a battery of backends will include `array-api-strict` in it, which does not contain any of the helper functions in its namespace.
So this would exclusively cause a regression for people who very specifically support only the four backends wrapped by `array-api-compat` and nothing else.

tests/test_common.py
```diff
@@ -276,7 +276,7 @@ def test_asarray_copy(library):
     is_lib_func = globals()[is_array_functions[library]]
     all = xp.all if library != 'dask.array' else lambda x: xp.all(x).compute()
 
-    if library == 'numpy' and xp.__version__[0] < '2' and not hasattr(xp, '_CopyMode') :
+    if library == 'numpy' and xp.__version__[0] < '2' and not hasattr(np, "_CopyMode"):
```
crusaderky, line 279: FYI this is for numpy 1.21

array_api_compat/common/_linalg.py
ev-br: Yeah, `test_all` is a bit brittle. At a guess, its iterating over `sys.modules` is meant to be robust against renames/new private submodules in `array_api_compat/*`. But then it's not easy to know if something was or was not imported.

ev-br (approved):
Looks good. +1 for not needing a deprecation layer. Thanks @crusaderky

PR #303, opened 2025-04-10
MAINT: simplify `torch` dtype promotion

crusaderky:
Salvaged from #298

- Simplify dtype promotion table
- Change `sum(x, axis=(), dtype=x.dtype)` and `prod` with the same parameters, which were previously returning `x` itself, to return a deep copy instead. While this is not part of the Array API, it's what both numpy an base torch do, and probably the most healthy thing to do xref https://data-apis.org/array-api/latest/design_topics/copies_views_and_mutation.html.
Discussion here: https://github.com/data-apis/array-api/issues/921

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -151,13 +151,18 @@ def result_type(
         return _reduce(_result_type, others + scalars)
 
 
-def _result_type(x, y):
+def _result_type(
+    x: Array | DType | bool | int | float | complex,
+    y: Array | DType | bool | int | float | complex,
+) -> DType:
     if not (isinstance(x, _py_scalars) or isinstance(y, _py_scalars)):
-        xdt = x.dtype if not isinstance(x, torch.dtype) else x
-        ydt = y.dtype if not isinstance(y, torch.dtype) else y
+        xdt = x if isinstance(x, torch.dtype) else x.dtype
+        ydt = y if isinstance(y, torch.dtype) else y.dtype
 
-        if (xdt, ydt) in _promotion_table:
+        try:
             return _promotion_table[xdt, ydt]
+        except KeyError:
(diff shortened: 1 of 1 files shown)


array_api_compat/torch/_aliases.py
```diff
@@ -35,54 +35,33 @@
...
     # complexes
-    (torch.complex64, torch.complex64): torch.complex64,
     (torch.complex64, torch.complex128): torch.complex128,
-    (torch.complex128, torch.complex64): torch.complex128,
-    (torch.complex128, torch.complex128): torch.complex128,
     # Mixed float and complex
     (torch.float32, torch.complex64): torch.complex64,
     (torch.float32, torch.complex128): torch.complex128,
     (torch.float64, torch.complex64): torch.complex128,
     (torch.float64, torch.complex128): torch.complex128,
```
crusaderky, line 59: `(complex, float)` use cases were missing

array_api_compat/torch/_aliases.py
```diff
@@ -301,27 +285,35 @@ def _reduce_multiple_axes(f, x, axis, keepdims=False, **kwargs):
...
+    if dtype is not None:
+        return x.clone() if dtype == x.dtype else x.to(dtype)
+
+    # We can't upcast uint8 according to the spec because there is no
+    # torch.uint64, so at least upcast to int64 which is what prod does
+    # when axis=None.
+    if x.dtype in (torch.uint8, torch.int8, torch.int16, torch.int32):
+        return x.to(torch.int64)
+
+    return x.clone()
```
ev-br, line 304: Note to self: this looks scary, but is in fact just a refactoring. Previously this stanza was duplicated in `sum` and `prod`.

ev-br (approved):
LGTM, thanks @crusaderky

PR #313, opened 2025-04-18
Type annotations, part 4

crusaderky:


array_api_compat/common/_helpers.py
```diff
@@ -12,56 +12,51 @@
...
     import torch
 
     # TODO: import from typing (requires Python >=3.13)
-    from typing_extensions import TypeIs, TypeVar
-
-    _SizeT = TypeVar("_SizeT", bound = int | None)
+    from typing_extensions import TypeIs
 
     _ZeroGradientArray: TypeAlias = npt.NDArray[np.void]
-    _CupyArray: TypeAlias = Any  # cupy has no py.typed
```
crusaderky, line 47: neither does dask...

array_api_compat/common/_helpers.py
```diff
@@ -280,13 +275,13 @@ def is_pydata_sparse_array(x: object) -> TypeIs[sparse.SparseArray]:
...
 
-    import sparse  # pyright: ignore[reportMissingTypeStubs]
+    import sparse
 
     # TODO: Account for other backends.
     return isinstance(x, sparse.SparseArray)
 
 
-def is_array_api_obj(x: object) -> TypeIs[_ArrayApiObj]:  # pyright: ignore[reportUnknownParameterType]
+def is_array_api_obj(x: object) -> TypeGuard[_ArrayApiObj]:
```
crusaderky, line 284: Given the current definition of _ArrayApiObj, TypeIs would cause downstream failures for all unknown array api compliant libraries.

array_api_compat/common/_typing.py
```diff
@@ -70,72 +69,11 @@ def shape(self, /) -> _T_co: ...
 DTypeKind: TypeAlias = _DTypeKind | tuple[_DTypeKind, ...]
 
 
-# `__array_namespace_info__.dtypes(kind="bool")`
```
crusaderky, line 73: holy overengineering Batman!

array_api_compat/numpy/__init__.py
```diff
@@ -23,13 +23,6 @@
...
 
-from ..common._helpers import *  # noqa: F403
-from .linalg import matrix_transpose, vecdot  # noqa: F401
-
-try:
-    # Used in asarray(). Not present in older versions.
-    from numpy import _CopyMode  # noqa: F401
-except ImportError:
-    pass
+from .linalg import matrix_transpose, vecdot  # type: ignore[no-redef] # noqa: F401
```
crusaderky, line 26: reverts regression from #288

array_api_compat/numpy/_aliases.py
```diff
@@ -92,7 +77,7 @@ def asarray(
     *,
     dtype: DType | None = None,
     device: Device | None = None,
-    copy: _Copy | None = None,
+    copy: py_bool | None = None,
```
crusaderky, line 80: Signature for Array API is bool.
The use of np._CopyMode is an internal of numpy 1 which should not be supported in annotations.

array_api_compat/cupy/_aliases.py
```diff
@@ -69,18 +68,13 @@
...
+    obj: Array | complex | NestedSequence[complex] | SupportsBufferProtocol,
     /,
     *,
-    dtype: Optional[DType] = None,
-    device: Optional[Device] = None,
-    copy: Optional[bool] = _copy_default,
-    **kwargs,
+    dtype: DType | None = None,
+    device: Device | None = None,
+    copy: py_bool | None = _copy_default,
```
crusaderky, line 76: This is plain wrong and requires a dedicated follow-up

crusaderky: https://github.com/data-apis/array-api-compat/pull/314

array_api_compat/common/_typing.py
```diff
@@ -26,13 +25,13 @@ def __getitem__(self, key: int, /) -> _T_co | NestedSequence[_T_co]: ...
     def __len__(self, /) -> int: ...
 
 
-class SupportsArrayNamespace(Protocol[_T_co]):
-    def __array_namespace__(self, /, *, api_version: str | None) -> _T_co: ...
+class SupportsArrayNamespace(Protocol):
+    def __array_namespace__(self, /, *, api_version: str | None) -> Namespace: ...
```
jorenham, line 29: There's a good reason why I did this, and I explained it two comments below.

How about we instead set the default of `_T_co` to `Namespace`? Then we're both happy.

array_api_compat/common/_helpers.py
```diff
@@ -956,7 +951,7 @@ def is_writeable_array(x: object) -> bool:
     return is_array_api_obj(x)
 
 
-def is_lazy_array(x: object) -> bool:
+def is_lazy_array(x: object) -> TypeGuard[_ArrayApiObj]:
```
jorenham, line 954: Anything with a `shape: tuple` that contains a `None` would return `True` here, so this isn't correct.

array_api_compat/common/_helpers.py
```diff
@@ -914,12 +909,12 @@ def to_device(x: Array, device: Device, /, *, stream: int | Any | None = None) -
...
-def size(x: HasShape[Collection[SupportsIndex]]) -> int: ...
+def size(x: HasShape[int]) -> int: ...
 @overload
-def size(x: HasShape[Collection[None]]) -> None: ...
+def size(x: HasShape[int | None]) -> int | None: ...
 @overload
-def size(x: HasShape[Collection[SupportsIndex | None]]) -> int | None: ...
-def size(x: HasShape[Collection[SupportsIndex | None]]) -> int | None:
+def size(x: HasShape[float]) -> int | None: ...  # Dask special case
+def size(x: HasShape[float | None]) -> int | None:
```
jorenham, line 917: why did you remove the overload that returns `-> None`?

array_api_compat/numpy/__init__.py
jorenham: what was the issue?

array_api_compat/numpy/_info.py
```diff
@@ -139,7 +141,7 @@ def default_dtypes(
         self,
         *,
         device: Device | None = None,
-    ) -> dict[str, dtype[intp | float64 | complex128]]:
+    ) -> DefaultDTypes:
```
jorenham, line 144: This is a lot less precise, so why did you change it?

pyproject.toml
```diff
@@ -79,20 +79,44 @@ ignore = [
...
+warn_unreachable = true
+
+[[tool.mypy.overrides]]
+module = ["cupy.*", "dask.*", "jax.*", "ndonnx.*", "sparse.*", "torch.*"]
+ignore_missing_imports = true
+
+
+[tool.pyright]
+include = ["src", "tests"]
+pythonVersion = "3.10"
```
jorenham, line 103: this isn't needed

```suggestion
```

it also makes type-checking in a CI matrix of different python version pretty much useless

pyproject.toml
```diff
@@ -79,20 +79,44 @@ ignore = [
...
-]
 
-ignore = [
-  # Module import not at top of file
-  "E402",
-  # Do not use bare `except`
-  "E722"
+[tool.mypy]
+files = ["array_api_compat"]
+python_version = "3.10"
```
jorenham, line 85: the comment below also applies here
```suggestion
```

pyproject.toml
```diff
@@ -79,20 +79,44 @@ ignore = [
...
+ignore_missing_imports = true
+no_implicit_optional = true
+show_error_codes = true
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_unreachable = true
+
+[[tool.mypy.overrides]]
+module = ["cupy.*", "dask.*", "jax.*", "ndonnx.*", "sparse.*", "torch.*"]
+ignore_missing_imports = true
```
jorenham, line 98: this is already `true` globally

jorenham (changes requested):
Many (if not most) of the changes here revert work that I've done in #288. Why didn't you address those while it was still open? Then I could've explained my reasons for it, which would have saved us both a lot time.

Seeing large portions of my work being deleted with comments like "holy overengineering Batman!" is pretty frustrating, to say the least. Because this way you're literally wasting the time and effort I've put into it, without even knowing *why* I made those decisions.

crusaderky:
> Why didn't you address those while it was still open? Then I could've explained my reasons for it, which would have saved us both a lot time.

I couldn't review that PR in time, apologies.

> Seeing large portions of my work being deleted with comments like "holy overengineering Batman!" is pretty frustrating, to say the least. 

I apologise, that was an emotional comment and I should not have made it.

array_api_compat/common/_aliases.py
```diff
@@ -720,7 +722,7 @@ def iinfo(type_: DType | Array, /, xp: Namespace) -> Any:
     "finfo",
     "iinfo",
 ]
-_all_ignore = ["inspect", "array_namespace", "NamedTuple"]
+_all_ignore = ["is_cupy_namespace", "inspect", "array_namespace", "NamedTuple"]
```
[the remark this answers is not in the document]
crusaderky: Good catch; I'll revert this one and write a separate PR that gets rid of `_all_ignore

array_api_compat/common/_aliases.py
```diff
@@ -381,8 +383,8 @@ def clip(
     # TODO: np.clip has other ufunc kwargs
     out: Array | None = None,
 ) -> Array:
-    def _isscalar(a: object) -> TypeIs[int | float | None]:
-        return isinstance(a, (int, float, type(None)))
+    def _isscalar(a: object) -> TypeIs[float | None]:
+        return isinstance(a, int | float | NoneType)
```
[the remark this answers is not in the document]
crusaderky: Why do we avoid the `types` import?

array_api_compat/common/_helpers.py
crusaderky: `SupportsArrayNamespace` just states that the object has a `__array_namespace__` method, nothing more.
It's missing all the other methods and properties of an Array.
I would much rather NOT write the full Array protocol here (I did it in array-api-extra and I regret it), as this is squarely in scope for `array-api-types`.

array_api_compat/common/_typing.py
crusaderky: I do not see much benefit at all in breaking down which dict keys each of the methods of `dtypes()` can return. From both a maintainer and a end user perspective, I can see very little to no gain for any work in addition to `-> Mapping[str, DType]`.

array_api_compat/common/_aliases.py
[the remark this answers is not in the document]
crusaderky: Ah. I hadn't noticed `__dir__`. It's a nice idea but it completely neuters `test_all`.
I will revert the changes here and write a separate PR to address the issue.

array_api_compat/common/_helpers.py
```diff
@@ -624,14 +619,14 @@ def your_function(x, y):
...
-                    import jax.experimental.array_api as jnp  # pyright: ignore[reportMissingImports]
+                    import jax.experimental.array_api as jnp  # type: ignore[no-redef]
             namespaces.add(jnp)
         elif is_pydata_sparse_array(x):
             if use_compat is True:
                 _check_api_version(api_version)
                 raise ValueError("`sparse` does not have an array-api-compat wrapper")
             else:
-                import sparse  # pyright: ignore[reportMissingTypeStubs]
+                import sparse
```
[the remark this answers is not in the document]
crusaderky: I'd rather not enlarge the scope of this PR.
On a side note, this is somewhere I disagree with the ruff policy. The else clause, while not being functionally useful, makes the code flow more readable IMHO.

array_api_compat/common/_helpers.py
```diff
@@ -12,56 +12,51 @@
...
-        | _CupyArray
+        | SupportsArrayNamespace
     )
 
 _API_VERSIONS_OLD: Final = frozenset({"2021.12", "2022.12", "2023.12"})
 _API_VERSIONS: Final = _API_VERSIONS_OLD | frozenset({"2024.12"})
 
 
-def _is_jax_zero_gradient_array(x: object) -> TypeGuard[_ZeroGradientArray]:
+def _is_jax_zero_gradient_array(x: object) -> TypeIs[_ZeroGradientArray]:
```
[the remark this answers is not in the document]
crusaderky: reverting

array_api_compat/common/_helpers.py
```diff
@@ -640,9 +635,9 @@ def your_function(x, y):
                 raise ValueError(
                     "The given array does not have an array-api-compat wrapper"
                 )
-            x = cast("SupportsArrayNamespace[Any]", x)
+            x = cast(SupportsArrayNamespace, x)
```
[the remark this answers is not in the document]
crusaderky: I _strongly_ disagree with TC006

array_api_compat/common/_helpers.py
```diff
@@ -776,12 +771,12 @@ def device(x: _ArrayApiObj, /) -> Device:
...
 # Based on cupy.array_api.Array.to_device
 def _cupy_to_device(
-    x: _CupyArray,
+    x: cp.ndarray,
     device: Device,
     /,
     stream: int | Any | None = None,
-) -> _CupyArray:
-    import cupy as cp  # pyright: ignore[reportMissingTypeStubs]
+) -> cp.ndarray:
```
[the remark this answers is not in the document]
crusaderky: Not that I'm aware of. But we have to skip all missing imports anyway or people that can't pip-install cupy won't be able to run the linter.

array_api_compat/common/_helpers.py
```diff
@@ -814,7 +809,7 @@ def _cupy_to_device(
             arr = x.copy()
         finally:
             runtime.setDevice(prev_device)  # pyright: ignore[reportUnknownMemberType]
-            if stream is not None:
+            if prev_stream is not None:
```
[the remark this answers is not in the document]
crusaderky: It's actually logically identical to before. But it was convoluted and rightfully the type checker was complaining.

array_api_compat/common/_helpers.py
```diff
@@ -823,7 +818,7 @@ def _torch_to_device(
     x: torch.Tensor,
     device: torch.device | str | int,
     /,
-    stream: None = None,
+    stream: int | Any | None = None,
```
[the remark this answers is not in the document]
crusaderky: there isn't one, and IMHO it's not referenced enough to warrant one

array_api_compat/common/_typing.py
```diff
@@ -26,13 +25,13 @@ def __getitem__(self, key: int, /) -> _T_co | NestedSequence[_T_co]: ...
...
 
 
 class HasShape(Protocol[_T_co]):
     @property
-    def shape(self, /) -> _T_co: ...
+    def shape(self, /) -> tuple[_T_co, ...]: ...
```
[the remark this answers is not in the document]
crusaderky: What information are we losing? Where would it be useful?

array_api_compat/common/_helpers.py
crusaderky: That's an issue with the implementation, not the type annotation.
As per the docstring, this function returns True if and only if x is an array API object and it is lazy.

array_api_compat/common/_helpers.py
crusaderky: Because no Array API backend in existence, present or future, will ever have _all_ its arrays with no shape information whatsoever.

pyproject.toml
crusaderky: removed

array_api_compat/common/_typing.py
crusaderky: I'm not sure I understand what the problem is; could you make an example?

crusaderky:
@jorenham can we get some traction back into this?

jorenham:
> @jorenham can we get some traction back into this?

I'll take a look tonight

array_api_compat/common/_linalg.py
```diff
@@ -8,7 +8,7 @@
 if np.__version__[0] == "2":
     from numpy.lib.array_utils import normalize_axis_tuple
 else:
-    from numpy.core.numeric import normalize_axis_tuple
+    from numpy.core.numeric import normalize_axis_tuple  # type: ignore[no-redef]
```
jorenham, line 11: Mypy's `allow_redefinition` will be enabled by default in mypy 2. So perhaps we could enable it project-wide? I believe that this ignore also wouldn't be necessary anymore.

array_api_compat/common/_typing.py
jorenham: So the `Namespace` that is a goldmine of static typing information. For instance, it implements the `dtypes` method ([docs](https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.dtypes.html#array_api.info.dtypes)), which can be used to extract the dtype *type* of an array-api object. A very rough example of that could look as follows:

```py
class ArrayNamespaceInfo[DTypeT](Protocol):
    def dtypes(self, *, device: Any = ..., kind: Any = ...) -> dict[Any, DTypeT]: ...

def get_dtype[DTypeT](
    a: SupportsArrayNamespace[ArrayNamespaceInfo[DTypeT]],
    /,
) -> DTypeT: ...
```

Using similar structural-typing trickery, it is also possible to extract the device type.
And if the library annotates its `dtypes` methods so that the return type depends on the `kind` literal passed to it (as I've tried to do here), then you would be able to distinguish between e.g. boolean and integer dtypes. That in turn allows you to write functions that reject booleans, but accept integers.

array_api_compat/common/_typing.py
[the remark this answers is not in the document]
jorenham: It's a bit of a work in progress right now, but I'm working an implementing shape-typing in numpy at the moment. But the python type system is pretty limited, so some tricks are needed to make shape-typing actually useful. 

My approach is to create `@type_check_only` subtypes of several (integer) `tuple`'s, e.g. `Rank0 <: tuple[()]`, `Rank2 <: tuple[int, int]`, and `Rank1N <: tuple[int, *tuple[int, ...]]` (1-d or more).
I then attach some non-existent (type-check-only) methods onto those `Rank` subtypes, that tell you which rank types are *smaller*, and which ones are *larger* than itself. It might sound trivial, but this information isn't possible to statically deduce from the `tuple` types themselves. When broadcasting arrays, this information is essential for determining what the output shape type will be. And broadcasting is *everywhere* within numpy, so shape-typing would be pretty useless without it.

Here's the current implementation btw (which is probably going to change a lot): https://github.com/numpy/numtype/blob/d7dcd8d6114cdbc56e0d43b9b834e3a49d1d147a/src/_numtype/_rank.pyi#L82-L137

But perhaps the tests might be more interesting: https://github.com/numpy/numtype/blob/d7dcd8d6114cdbc56e0d43b9b834e3a49d1d147a/src/_numtype/%40test/generated/test_rank.pyi

> What information are we losing? Where would it be useful?

To make a long story short: shape-typing in numpy

array_api_compat/common/_typing.py
jorenham: Two comments above, I explain how you can determine the type of e.g. boolean and integer dtypes, and how this batmobile is the (only) way of getting there. With this, it becomes possible to annotate an array-api function that rejects boolean arrays, yet accepts integer arrays.

array_api_compat/numpy/_info.py
jorenham: > The ultra-precise annotation was giving neither final users nor maintainers any benefit.

Users can always decide to broaden the returned type themselves, using their own annotations. But the other way around requires nasty things like `cast`. 
But I don't feel very strongly about this one, so I'll stop myself from ranting about a bunch of other benefits....

> Why did you define a TypeAlias if you're not using it?

If it's not needed, then feel free to remove it. But using it just because it exists feels a bit backwards to me.

array_api_compat/common/_aliases.py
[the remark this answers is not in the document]
crusaderky: done

array_api_compat/common/_helpers.py
crusaderky: Today, yes, because `array-api-compat` by design does not wrap around Array classes. NumPy 1.x, Dask, CuPy, and PyTorch don't have this method even if it's mandated by the Array API standard.
`array-api-compat` shims around this by defining the `is_array_api_obj` function, which states that an Array API array is one defined by one of the four above libraries _or by any fully compliant library_.

array_api_compat/common/_helpers.py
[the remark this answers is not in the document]
crusaderky: mypy and pyright must work without `cupy` installed. This is because `cupy` cannot be installed on hosts without CUDA hardware, and is not available on MacOS regardless of hardware. As there is no `cupy-types` stubs package, the only sensible solution is to just flat out suppress it with `ignore_missing_imports = True` (see `pyproject.toml`).

`torch` has the same issue, to a lesser extent. While it is available on all architectures, it is _really really_ heavy, to the point that it's unsavory to install it in a pre-commit environment. 

Ditto for `jax`, which is very heavy and also unavailable on Windows.

`array-api-extra` already skips all three for this reason in its `lint` environment.

array_api_compat/common/_linalg.py
crusaderky: Adding `allow_redefinition = true` to pyproject.toml doesn't seem to have any effect in mypy 1.15.0

array_api_compat/common/_typing.py
crusaderky: I don't have the bandwidth to discuss this anymore. Reverting.

crusaderky:
@jorenham 
I've reverted my changes re. shape-typing, namespace generics, super-specific dtypes() return value annotations, plus a bunch of other minor things.
I believe there are no outstanding issues but I may have missed some.
Ready for another iteration of review.

array_api_compat/common/_helpers.py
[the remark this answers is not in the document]
jorenham: Ok thanks. We're using `reportMissingImports = false` now, so it's probably better this way 👌🏻.

jorenham (approved):
I'm all out of things to whine about. And my apologies for getting over-defensive. I appreciate your work :)

crusaderky:
@ev-br ready for merge

PR #314, opened 2025-04-18
ENH: Simplify CuPy `asarray` and `to_device`

crusaderky:
Salvaged from #293 
Simplify to_device and asarray for cupy and implement copy=False


array_api_compat/common/_helpers.py
```diff
@@ -781,42 +781,28 @@ def _cupy_to_device(
...
-    elif device == "cpu":
+    if device == "cpu":
         # allowing us to use `to_device(x, "cpu")`
         # is useful for portable test swapping between
         # host and device backends
         return x.get()
-    elif not isinstance(device, _Device):
-        raise ValueError(f"Unsupported device {device!r}")
-    else:
-        # see cupy/cupy#5985 for the reason how we handle device/stream here
```
crusaderky, line 799: Issue fixed in 2021

array_api_compat/common/_helpers.py
```diff
@@ -781,42 +781,28 @@ def _cupy_to_device(
...
+    import cupy as cp
 
-    if device == x.device:
-        return x
-    elif device == "cpu":
+    if device == "cpu":
         # allowing us to use `to_device(x, "cpu")`
         # is useful for portable test swapping between
         # host and device backends
         return x.get()
```
ev-br, line 790: Whoa. I'm confused: this returns a numpy array if `x` was a cupy array. This might be intended, but then the return annotation is misleading? Same for the device argument annotation. So maybe add a comment if a numpydoc docstring with `Parameters` and `Returns` sections is an overkill.

array_api_compat/common/_helpers.py
```diff
@@ -781,42 +781,28 @@ def _cupy_to_device(
...
+            return cp.asarray(x)
+
+    # stream can be an int as specified in __dlpack__, or a CuPy stream
+    if isinstance(stream, int):
+        stream = cp.cuda.ExternalStream(stream)
+    elif not isinstance(stream, cp.cuda.Stream):
+        raise TypeError(f"Unsupported stream type {stream!r}")
+
+    with device, stream:
+        return cp.asarray(x)
```
ev-br, line 805: Do we have test coverage for this part? TBH I'm ready to believe it's totally correct but would prefer to have some tests to ensure that it really is.
Also I wouldn't be totally surprised if this is cupy version dependent, in which case a decent test coverage would be really helpful for diagnosing/debugging.

array_api_compat/cupy/_aliases.py
```diff
@@ -89,25 +87,13 @@ def asarray(
...
-            kwargs['copy'] = copy
-
-        return cp.array(obj, dtype=dtype, **kwargs)
+        if copy is None:
+            return cp.asarray(obj, dtype=dtype, **kwargs)
+        else:
+            res = cp.array(obj, dtype=dtype, copy=copy, **kwargs)
+            if not copy and res is not obj:
+                raise ValueError("Unable to avoid copy while creating an array as requested")
+            return res
```
ev-br, line 96: This part seems to streamline and remove workarounds for presumably older cupy versions. 

1. Do we have test coverage?
2. Which cupy version made the change which renders the removed workaround obsolete?
3. Do we then need to formulate some sort of policy for a minimum supported CuPy version? I personally would be OK with any sort of policy ("the currently most recent released version" is fine by me), as long as it's spelled out and lower capped in `pyproject.toml`.

ev-br (changes requested):
Several questions inline.
I think it needs a bit of clarification on test coverage and cupy version support.

array_api_compat/common/_helpers.py
crusaderky: None of this has changed. See also #87.

array_api_compat/cupy/_aliases.py
crusaderky: Nope, the latest cupy still has the same exact issues as the old ones.
The API I'm using hasn't changed since CuPy 1.0.

> Do we then need to formulate some sort of policy for a minimum supported CuPy version? 

We need to formulate policy for a minimum supported version of _all backends_.
So far it has not been done deliberately in order to reduce attrition (much appreciated).
This is a matter of discussion for the Committee.

ev-br: The latest cupy you mean 13.4 or 14.0.a1? 

One thing to bear in mind with CuPy specifically: IIUC they are still working through the numpy 2.0 transition. 13.4 is sort-of compatible (one goal was to make it import with both numpy 1 and numpy 2), 14.x series is supposed to complete the transition.
Which is partly a motivation for my questions/comments: if it's a moving target now, we should have test coverage.

array_api_compat/common/_helpers.py
ev-br: Yes, testing it in array-api-tests would be great. Two things to consider though:

- Given that there are no tests at all, either here (cupy-specific) or generic, I think this PR is a great opportunity to add some testing
- Given how underspecified the device story is in the standard (for better or worse), my impression is that array-api-tests would be somewhat limited (as we cannot test implementation-dependent behaviors in there). Thus, cupy-specific tests here would be great.

Re: how old is a version we care? Might want some input from CuPy devs, too. Off the cuff, I'd say we definitely care about 13.4; previous 13.x versions with `x < 4`, not necessary; 
Whether we care about 13.x when 14.x stops being alpha... Maybe not unless there are strong reasons?

array_api_compat/common/_helpers.py
ev-br: My point is that gh-87 is still open, and there are no tests, so if we touch this, great, let's take the opportunity to also improve testing.

array_api_compat/cupy/_aliases.py
crusaderky: 13.4.1.

crusaderky:
@ev-br added unit tests; all green locally on CuPy 13.4.1. There should not be any outstanding complaints.

tests/test_common.py
```diff
@@ -189,23 +190,26 @@ class C:
...
 
-    x2 = to_device(x, dev)
-    assert device(x2) == device(x)
-
-    x3 = xp.asarray(x, device=dev)
-    assert device(x3) == device(x)
+    for dev in devices:
+        if dev is None:  # JAX >=0.5.3
+            continue
+        if dev is _DASK_DEVICE:  # TODO this needs a better design
```
ev-br, line 209: Minor, and probably I'm just being dense: what needs better design: the test, _DASK_DEVICE, checking for it?

tests/test_cupy.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def test_to_device_with_stream():
+    devices = xp.__array_namespace_info__().devices()
+    streams = [
+        Stream(),
+        Stream(non_blocking=True), 
+        Stream(null=True),
+        Stream(ptds=True), 
+        123,  # lapack stream
```
ev-br, line 15:
```suggestion
        123,  # dlpack stream
```

typo? unlikely that linear algebra is anyhow related

ev-br (commented):
LGTM modulo a question and a nit.

tests/test_common.py
crusaderky: `_DASK_DEVICE` is a problematic design. I'll try to improve it a little in a follow-up PR.

ev-br:
Merged, thanks @crusaderky

PR #319, opened 2025-05-09
MAINT: update CuPy xfails

ev-br:


ev-br:
Merging straight away: the main push of this PR is to update the CuPy xfails, which we do not test on CI anyway. The numpy==1.22 back-and-forth is just jitter.

PR #321, opened 2025-05-11
WIP: Test CI speedups

ev-br:
testing, do not merge: this checks out array-api-tests from a branch of  a fork

The matching array-api-tests PR, https://github.com/data-apis/array-api-tests/pull/373, allows to turn xfails into skips. 
Apparently, this has a large effect on run time: in fact the vast majority of time on CI was spent in hypothesis working to distill a failure---for things which we know are failing and have explicitly marked as xfails.


ev-br:
Okay, this is an obvious win.
PyTorch job used to take ~6min, in this PR it is 2m14s
Dask job used to take ~13m30s, in this PR it is ~40s
NumPy wins are marginal, down to just under 2 minutes from just above 2 minutes.

The largest benefactor is Dask: we had to limit the `--max-examples` to 5, meaning we can up the number to ~100. Still in the "smoke testing"  realm, but still closer to what other backends are using.

ev-br:
The last commit bumps the number of examples to 500 for NumPy and PyTorch (up from 200) and to 50 for Dask (up from 5). All run finish in under 5 minutes, which is I think is a sweet spot for the balance between the convenience of iterating on PR and a reasonable test coverage.

ev-br:
Using 4 pytest workers, `-n 4`, allows to bump the number of examples to 1000 for numpy/pytorch and 200 for dask; run time is still under 5-10 minutes (closer to 6-7 for torch, under 5 for others). I suppose we'll need to keep an eye on how this fluctuates but overall this sounds like a win.

One potential point of concern can be how parallelization affects the failure mode. It's already not always easy to untangle how hypothesis reports errors and replicate a CI failure locally. I suppose we'll try and see if parallelization makes it any more difficult.

ev-br:
Okay, I think this is it for low-hanging fruits. The slowest tests on torch and numpy spend their time inside hypothesis, so not much we can do about it.
For dask (and presumably, jax and cupy), further improvements are possible, with significantly more drastic changes. Likely using some ideas from https://github.com/data-apis/array-api-tests/issues/197 .

ev-br:
cc @crusaderky this seems to partially address some of your concerns about testing dask

ev-br:
Now that https://github.com/data-apis/array-api-tests/pull/373 has landed, array-api-tests are checked out from the master branch again, and the CI improvements are still there (https://github.com/data-apis/array-api-compat/pull/321#issuecomment-2872434210). Failure reports are still as readable as before, so this seems to be a win all around. Merging.

PR #322, opened 2025-05-11
BUG: torch: meshgrid defaults to indexing="xy"

ev-br:
As of version 2.6, torch defaults to indexing='ij', and is planning to transition to 'xy' at some point. When it does, we'll be able to drop our wrapper.

cross-ref https://github.com/pytorch/pytorch/issues/50276 for torch-specific details.

One other difference from the spec is that pytorch version returns a tuple of arrays, which matches numpy but deviates from jax or cupy, cross-ref https://github.com/data-apis/array-api/issues/938

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from functools import reduce as _reduce, wraps as _wraps
 from builtins import all as _builtin_all, any as _builtin_any
-from typing import Any, List, Optional, Sequence, Tuple, Union
+from typing import Any, List, Optional, Sequence, Tuple, Union, Literal
 
 import torch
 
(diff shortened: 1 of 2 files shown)


ev-br:
(facepalm). Will send a fix in a while unless someone beats me to it.

array_api_compat/torch/_aliases.py
```diff
@@ -828,6 +828,12 @@ def sign(x: Array, /) -> Array:
         return out
 
 
+def meshgrid(*arrays: Array, indexing: Literal['xy', 'ij'] = 'xy') -> List[Array]:
+    # enforce the default of 'xy'
+    # TODO: is the return type a list or a tuple
+    return list(torch.meshgrid(*arrays, indexing='xy'))
```
mdhaber, line 834:
```sugestion
    return list(torch.meshgrid(*arrays, indexing=indexing))
```

mdhaber:
I would do it, but I don't know why it's not caught by tests? Are they not running for torch?

ev-br:
Thanks! The test this PR added is just too dumb, and only checks that the default is "xy"

PR #325, opened 2025-05-15
TST: revisit test for `asarray` `copy=` parameter

### LINKED ISSUE #324: Test errors with cupy==13.4.1
```
FAILED tests/test_common.py::test_asarray_copy[cupy] - Failed: DID NOT RAISE <class 'NotImplementedError'>
FAILED tests/test_cupy.py::test_to_device_with_stream - RuntimeError: The device that the stream is created on (0) does not match with the...
```

This is probably related to https://github.com/data-apis/array-api-compat/pull/314 @crusaderky 

EDIT: and indeed, `$ git revert  e600449` makes tests pass again.

crusaderky:
- Partially fixes #324
- Dask now will raise `ValueError` instead of `NotImplementedError` when it fails `asarray(..., copy=False)`.
- Align test to reflect changes in #314

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -278,87 +278,73 @@ def test_asarray_copy(library):
     xp = import_(library, wrapper=True)
     asarray = xp.asarray
     is_lib_func = globals()[is_array_functions[library]]
-    all = xp.all if library != 'dask.array' else lambda x: xp.all(x).compute()
-
-    if library == 'cupy':
-        supports_copy_false_other_ns = False
-        supports_copy_false_same_ns = False
-    elif library == 'dask.array':
-        supports_copy_false_other_ns = False
-        supports_copy_false_same_ns = True
-    else:
-        supports_copy_false_other_ns = True
-        supports_copy_false_same_ns = True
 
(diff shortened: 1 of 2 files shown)


tests/test_common.py
```diff
@@ -278,100 +277,87 @@ def test_asarray_copy(library):
...
     a[0] = 0.0
-    assert all(b[0] == 1.0)
+    assert b[0] == 1.0
 
-    a = array.array('f', [1.0])
-    if supports_copy_false_other_ns:
+    a = array.array("f", [1.0])
+    if library in ("cupy", "dask.array"):
+        with pytest.raises(ValueError):
+            asarray(a, copy=False)
```
ev-br, line 341: Can't say I'm happy about the need for branching here. It is what it is though

ev-br (approved):
Thanks Guido. Let's give it a shot, and test it locally together with other fixes.

tests/test_common.py
crusaderky: You're not happy that cupy can't create a GPU array as a view of host memory? 😕

PR #373, opened 2026-01-06
BLD/DEV: use meson-python, add a basic Pixi workspace

lucascolley:
Clearly plenty more work to do to get this up to par with array-api-extra, but I think that merging something like this would be a good start.

Follow-ups:
- Convert CI to Pixi
- add linting

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-pixi.lock
 vendor_tests/array_api_compat
 
 # Byte-compiled / optimized / DLL files
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+pixi.lock
+vendor_tests/array_api_compat
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/docs/dev/releasing.md
+++ b/docs/dev/releasing.md
@@ -44,14 +44,13 @@
 
 - [ ] **Update the version.**
 
-  You must edit
+  You must edit the version in:
 
   ```
   array_api_compat/__init__.py
+  pyproject.toml
+  meson.build
   ```
-
-  and update the version (the version is not computed from the tag because
-  that would break vendorability).
 
 - [ ] **Update the [changelog](../changelog.md).**
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,10 +1,10 @@
 [build-system]
-requires = ["setuptools", "setuptools-scm"]
-build-backend = "setuptools.build_meta"
+requires = ["meson-python"]
+build-backend = "mesonpy"
 
 [project]
 name = "array-api-compat"
-dynamic = ["version"]
+version = "1.15.0.dev0"
 description = "A wrapper around NumPy and other array libraries to make them compatible with the Array API standard"
 readme = "README.md"
 requires-python = ">=3.10"
@@ -55,15 +55,10 @@ dev = [
 homepage = "https://data-apis.org/array-api-compat/"
 repository = "https://github.com/data-apis/array-api-compat/"
 
-[tool.setuptools.dynamic]
-version = { attr = "array_api_compat.__version__" }
 
-[tool.setuptools]
-package-dir = {"" = "src"}
+[tool.pytest.ini_options]
+testpaths = ["tests"]
 
-[tool.setuptools.packages.find]
-where = ["src"]
-namespaces = false
 
 [tool.ruff.lint]
 preview = true
--- a/vendor_tests/test_vendoring.py
+++ b/vendor_tests/test_vendoring.py
@@ -2,7 +2,7 @@ import pytest
 
 
 def test_vendoring_numpy():
-    from vendor_test import uses_numpy
+    from . import uses_numpy
 
     uses_numpy._test_numpy()
 
@@ -10,19 +10,20 @@ def test_vendoring_numpy():
 def test_vendoring_cupy():
     pytest.importorskip("cupy")
 
-    from vendor_test import uses_cupy
+    from . import uses_cupy
 
     uses_cupy._test_cupy()
 
 
 def test_vendoring_torch():
     pytest.importorskip("torch")
-    from vendor_test import uses_torch
+    from . import uses_torch
 
     uses_torch._test_torch()
 
 
 def test_vendoring_dask():
     pytest.importorskip("dask")
-    from vendor_test import uses_dask
+    from . import uses_dask
+
     uses_dask._test_dask()
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# SCM syntax highlighting & preventing 3-way merges
+pixi.lock merge=binary linguist-language=YAML linguist-generated=true -diff
--- a/.github/workflows/publish-package.yml
+++ b/.github/workflows/publish-package.yml
@@ -41,7 +41,7 @@ jobs:
 
     - name: Install python-build and twine
       run: |
-        python -m pip install --upgrade pip "setuptools<=67"
+        python -m pip install --upgrade pip meson-python
         python -m pip install build twine
         python -m pip list
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -66,3 +66,9 @@ jobs:
 
       - name: Run Tests
         run: pytest -v
+
+      - name: Run vendoring tests
+        run: |
+          rm -rf vendor_tests/array_api_compat
+          cp -r src/array_api_compat vendor_tests/
+          pytest -v vendor_tests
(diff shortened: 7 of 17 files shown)


pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "array-api-compat"
-dynamic = ["version"]
+version = "1.13.0"
```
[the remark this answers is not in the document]
ev-br: Not sure TBH. One one hand, the version is manually set in `__init__`, so it looks like we're not using dynamic versioning. Or maybe this "dynamic" setting in pyproject.toml makes it pick the version from `__init__`.
There's also a bit of magic around trusted publishing and I've no idea how this works (other than _push a tag and it just works_). 

I suppose the way to check it is make a change in a branch, then try making a release to a  testpypi (requires a manual editing of the relevant GH workflow) and see if this still works.

My personal prio for now is to get ready for 2025.12 spec release. That, among other things, requires a lot of manual testing locally, of matching branches of multiple projects. My workflow involves a lot of 
```
$ pushd ../array-api-tests
$ git co matching_branch
$ run tests
# swear quietly
$ popd 
# fix
$ pip install -e .
$ pushd ../array-api-tests
# run tests again
# swear loudly
# fix it in -tests
# repeat until convergence
``` 

not sure how that translates to the pixified workflow (swearing is optional), and I'd prefer to do a single thing at a time.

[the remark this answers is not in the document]
lucascolley: yeah, this could be a whole lot nicer with Pixi but it will need some hours of setup. I'll let you swear loudly for now :)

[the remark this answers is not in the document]
rgommers: Using an explicit version number is fine. Opinions obviously vary, but I'd consider it an improvement.

Versioning is unrelated to trusted publising.

ev-br (changes requested):
Let's maybe revive this if you're interested @lucascolley ? Sorry it took me long to get back to this.

I trust you on anything `pixi`. If dynamic versioning gets in the way, I'd be happy to see the back of it, static is indeed much nicer. 

What needs to be done if we change the versioning:
- purge any mentions of `setuptools_scm` and `tool.setuptools.dynamic` from `pyproject.toml`
- update the releasing docs to mention that the version needs to be bumped in both `__init__.py` and `pyproject.toml` it currently says "update the version in `__init__.py`"
- cut a dummy release locally (`python -m build . && twine check --strict dist/`
- make a dummy TestPyPI release to verify it all works

lucascolley:
I implemented the switch from setuptools to meson-python

ev-br (changes requested):
Agreed that the change to meson-python is nice. 

I actually wonder about vendoring scenarios. When e.g. SciPy vendors array-api-compat, it has to list the sources in the `meson.build` file. Now that `array-api-compat` has its meson.build file with the listing, can scipy reuse it? This is not blocking this PR of course, am just thinking out loud.

> make a dummy TestPyPI release to verify it all works

Thinking about it, to test trusted publishing it needs to be done from main/tag etc. Presumably trusted publishing is decoupled enough, so that it's more work than justified now. So checking that a release wheel looks reasonable locally is probably sufficient.

lucascolley:
Yes, this is the exact context within which I started this work: https://github.com/scipy/scipy/blob/main/meson.build#L216-L246

lucascolley:
@ev-br please could you let me know if you have any remaining concerns? thanks!

vendor_tests/test_vendoring.py
```diff
@@ -2,27 +2,28 @@
...
-    from vendor_test import uses_torch
+    from . import uses_torch
 
     uses_torch._test_torch()
 
 
 def test_vendoring_dask():
     pytest.importorskip("dask")
-    from vendor_test import uses_dask
+    from . import uses_dask
```
ev-br, line 27: Could you explain this and related changes please?

.gitignore
```diff
@@ -1,3 +1,6 @@
+pixi.lock
+vendor_tests/array_api_compat
```
ev-br, line 1: Could you explain these two? 

- ignoring `pixi.lock`: naively, I'd expect that if some dependency breaks, we'd want to bisect `pixi.lock` updates. From experience, is this not a useful workflow?

- Never saw any issues with `vendor_tests/array_api_compat` (or the very directory for that matter), why do we want to ignore it now?

docs/dev/releasing.md
```diff
@@ -44,14 +44,15 @@
 
 - [ ] **Update the version.**
 
-  You must edit
+  You must edit both of
```
ev-br: three places now: `__init__.py`, meson.build and pyproject.toml?

ev-br (commented):
A couple of questions below.
Other than that:
- pixi changes I happily defer to you Lucas :-).
- dynamic versioning -> static versioning change is great, let's rid of it.
- meson is a bit of an overkill for a pure python package --- I'd use flit for a pure-python package with static versioning, but OK if we want to use meson throughout;
- `src` layout: is it needed (why?) or is it just because why not (also why though?) 
- could you double-check what scipy and other vendorers need to do to adapt?

lucascolley:
> * meson is a bit of an overkill for a pure python package --- I'd use flit for a pure-python package with static versioning, but OK if we want to use meson throughout;

I agree that in general meson-python is overkill for a pure Python package, but in this specific case where we want to make it easily vendor-able by projects using meson-python, it makes sense

> * `src` layout: is it needed (why?) or is it just because why not (also why though?)

I think I originally introduced it because it kept `hatchling` on default settings happy. But indeed there are reasons why it is valuable, e.g. avoiding accidental use of relative imports. Let me know if you have more questions after reading https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/.

> * could you double-check what scipy and other vendorers need to do to adapt?

The work is already done for SciPy, this just enables us to delete https://github.com/scipy/scipy/blob/main/subprojects/array_api_compat/meson.build. Other vendorers can copy SciPy's approach.

.gitignore
lucascolley: I'm happy to commit `pixi.lock`. Without committing it, each individual contributor may have different environments on their machine, although they will all still be compatible with `pixi.toml`.

I was making the minimal change in case there were objections to committing the lock file, but if there aren't, then great, let's commit it.

lucascolley: > * Never saw any issues with `vendor_tests/array_api_compat` (or the very directory for that matter), why do we want to ignore it now?

The vendoring tests are now changed from the previous symlink at `vendor_test/vendored/_compat` to a `cp` command that populates this newly gitignored directory. We discussed above that we wanted to get rid of the symlink.

There is a Pixi task for this `cp` command that runs automatically with `pixi run tests-vendor`, and it is currently hardcoded in CI. I've marked converting CI to Pixi for a follow-up.

vendor_tests/test_vendoring.py
lucascolley: `test_vendoring` was moved from the `tests` directory to the `vendor_tests` directory, and this line is updated with the appropriate relative import. This matches what works well for us at https://github.com/data-apis/array-api-extra/tree/main/vendor_tests.

ev-br:
>> double-check what scipy and other vendorers need to do to adapt?

>The work is already done for SciPy, this just enables us to delete https://github.com/scipy/scipy/blob/main/subprojects/array_api_compat/meson.build. Other vendorers can copy SciPy's approach.

Great if it's just deleting a now extraneous source file.

Re `src` layout: I now read the packaging page, and TBH none of the stated reasons seem very compelling, in the sense that I don't remember seeing either of problems that pages warns of for non-src layouts.
So I agree with https://github.com/data-apis/array-api-compat/pull/373#discussion_r2961365631 : could you please undo the layout change here and submit it as a separate PR with a `.git-blame-ignore-revs` entry.

.gitignore
ev-br: Re commiting `pixi.lock` or not, I don't have an opinion and am happy to follow recommended best practices. If Lucas' recommendation is to not commit it, great, that settles it for me, thanks.

lucascolley:
> So I agree with #373 (comment) : could you please undo the layout change here and submit it as a separate PR with a .git-blame-ignore-revs entry.

Can we do the layout change first then rebase this? Should be less work overall.

ev-br:
(in case the "reaction" above does not generate a notification)

> do the layout change first then rebase this?

Sure!

lucascolley:
let me know if there is anything else!

lucascolley:
The CI failure looks unrelated

ev-br (approved):
LGTM now, modulo a small request before clicking the green button: could you reword the commit message of the last commit which brought in `pixi.lock`? Squash-merging I'd rather avoid as it's 3kLOC with the `pixi.lock` in.

lucascolley:
thanks for the review!

ev-br:
Thanks Lucas

PR #381, opened 2026-01-19
TST: update -xfails for complex special cases

ev-br:
New xfails for complex special cases, cf https://github.com/data-apis/array-api-tests/pull/409


ev-br:
All green on CI now. Merging to remove noise from unrelated PRs. Might want to report xfails to numpy/torch at some point.

PR #425, opened 2026-04-27
BUG: torch.meshgrid RuntimeError if no arrays

### LINKED ISSUE #424: torch.meshgrid raises RuntimeError if no arrays are passed
If no arrays are passed to `torch.meshgrid` a `RuntimeError` is raised whereas numpy returns an empty tuple.

```python
import torch
torch.meshgrid()
```
raises:
```
RuntimeError: meshgrid expects a non-empty TensorList
```
Whereas with numpy:
```python
import numpy as np
np.meshgrid()
```
returns:
```
()
```

I believe the numpy implementation is the correct output given the wording ("tuple of N arrays, where N is the number of provided one-dimensional input arrays") of the [docstring for meshgrid](https://data-apis.org/array-api/latest/API_specification/generated/array_api.meshgrid.html#meshgrid).

Nin17:
fixes https://github.com/data-apis/array-api-compat/issues/424

### DIFF
--- a/array_api_compat/torch/_aliases.py
+++ b/array_api_compat/torch/_aliases.py
@@ -936,7 +936,7 @@ def meshgrid(*arrays: Array, indexing: Literal['xy', 'ij'] = 'xy') -> tuple[Arra
     # torch <= 2.9 emits a UserWarning: "torch.meshgrid: in an upcoming release, it
     # will be required to pass the indexing argument."
     # Thus always pass it explicitly.
-    return torch.meshgrid(*arrays, indexing=indexing)
+    return torch.meshgrid(*arrays, indexing=indexing) if arrays else ()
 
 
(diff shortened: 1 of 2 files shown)


Nin17:
Should I also update the tests to test for this? Or should this be tested at array-api-tests?

ev-br:
> Should I also update the tests to test for this? Or should this be tested at array-api-tests?

Given that the [array-api-tests test](https://github.com/data-apis/array-api-tests/blob/55fcc60179efa2680ddd6cd926ddf17b83530e2b/array_api_tests/test_creation_functions.py#L547-L580) does not catch this, I think it could be nice to add a quick test here, yes. https://github.com/data-apis/array-api-compat/blob/main/tests/test_torch.py#L117 is the existing `torch` / `meshrid` test.

ev-br:
LGTM, merged. Thank you @Nin17 !

PR #428, opened 2026-05-27
MAINT: refactor directories

lucascolley:
for gh-373


.github/workflows/array-api-tests.yml
```diff
@@ -78,6 +78,9 @@ jobs:
         python -m pip install -r ${GITHUB_WORKSPACE}/array-api-tests/requirements.txt
         python -m pip install pytest-xdist
 
+    - name: Install array-api-compat
+      run: python -m pip install ${GITHUB_WORKSPACE}/array-api-compat
```
ev-br, line 82: This is now needed because of the `src` layout, while previously tests were picking up the sources from the source tree, correct?

pyproject.toml
```diff
@@ -56,10 +56,13 @@ homepage = "https://data-apis.org/array-api-compat/"
 repository = "https://github.com/data-apis/array-api-compat/"
 
 [tool.setuptools.dynamic]
-version = {attr = "array_api_compat.__version__"}
+version = { attr = "array_api_compat.__version__" }
```
ev-br, line 59: and this will go in gh-373?

.github/workflows/array-api-tests.yml
lucascolley: Yes, the non-obvious implicit source tree import

Issue #162, opened 2024-07-24
Add compatibility layer for Apple mlx


j-emberton:
MLX is an array library for use on apple silicon. It is intended to closely follow the python Array API however some difference exist.

https://github.com/ml-explore/mlx?tab=readme-ov-file

It would be useful to add compatibility layers for this app to support users wishing to do array operations on apple gpu

lucascolley:
x-ref ml-explore/mlx#48

asmeurer:
@j-emberton this is definitely in scope for array-api-compat. If you want to work on this, the best thing to do would be to fork this repository and make a pull request. 

Also note that partial support is OK to begin with, so if there's some function that's difficult to support, we can skip it for now. The development notes in the docs are a good place to start for contributing https://data-apis.org/array-api-compat/dev/index.html

asmeurer:
Also there has been some upstream work in mlx to support the array API (e.g., https://github.com/ml-explore/mlx/pull/1289), so you may want to synchronize with them to see what their plans are. If they are planning to add full support directly to mlx, that is preferable to adding support to the compat library. But if there are things they can't support easily because of backwards compatibility, then we may need a compat wrapper.

asmeurer:
Based on the discussion at https://github.com/ml-explore/mlx/pull/1289, the MLX folks are open to having full compatibility in MLX itself, so any work on upstream incompatibilities should go there. For now, let's just add helper functions for MLX to array-api-compat.

asmeurer:
I don't know if the upstream is doing anything here, but it might be worth revisiting this if they aren't. MLX is a useful library.

aaishwarymishra:
| Area | Array API expectation | MLX status today | Current approach in PR | Risk / open question | Priority |
|---|---|---|---|---|---|
| `device()` / `to_device()` | `device(to_device(x, dev)) == dev` | No real device ownership; unified memory | Virtual `id(array) -> device` registry | Is virtual device tracking acceptable long-term? Any GC / identity edge cases? | Blocker |
| Boolean indexing (`__getitem__`) | Boolean masks must work | Native MLX raises on boolean indexing | Monkeypatch `mx.array.__getitem__` and fallback to NumPy | Is global monkeypatching too invasive? Can this be limited or upstreamed? | Blocker |
| `float64` on GPU | Supported where backend allows, or clearly handled | Not supported on MLX GPU | Document as limitation / future upstream work | Does this block compliance or just specific tests? | Blocker |

[...]

PR #449, opened 2026-07-12
Mlx

aaishwarymishra:
### MLX Support Overview

  MLX is built from the ground up for Apple Silicon using a Unified Memory model—the CPU and GPU share the exact same physical memory. This presents
  unique design challenges when wrapping it for the Array API standard (which assumes distinct hardware platforms like CPU vs. GPU VRAM).

  In general,  array-api-compat  wraps existing libraries without subclassing or wrapping their array objects, keeping it lightweight. However, because
  MLX is missing a few core features required by the standard (like data-dependent shapes and explicit device ownership), we implemented a clean virtual-
  routing and monkeypatching solution:
  ──────
  ### The Monkeypatch: Boolean Indexing ( __getitem__ )

  #### 1. The Problem

  MLX is designed with an asynchronous graph compilation model where the shape of every array must be statically known at graph-build time.
  However, boolean indexing (e.g.  x[x < 5] ) is a data-dependent operation: the size of the returned array depends on the number of  True  values inside
  the mask, which isn't known until evaluation. Because of this, MLX arrays natively raise a  ValueError: boolean indices are not yet supported  when you
  try to index them.

  #### 2. The Solution

  Since the Array API standard guarantees that boolean indexing works, and standard routines (like  clip() ) rely on it, we monkeypatched the C++ array
  class's indexing method directly:

    # In src/array_api_compat/mlx/__init__.py
    _old_getitem = mx.array.__getitem__

    def _new_getitem(self, item):
        if _has_bool_array(item):
            # Convert array and boolean mask to numpy, index on CPU, and wrap back to MLX
            self_np = np.array(self)
            item_np = _to_numpy_indices(item)
            res_np = self_np[item_np]
            return mx.array(res_np)
        return _old_getitem(self, item)

    mx.array.__getitem__ = _new_getitem

  This intercepts indexing calls. If a boolean mask is detected, it falls back to CPU evaluation via NumPy and converts the result back to an MLX array.
  If no boolean mask is involved, it passes the call through to MLX's native indexing.

  (Note: We did not need to patch  __setitem__  because MLX natively supports in-place boolean assignments, e.g.  x[mask] = 1.0 .)
  ──────
  ### The Virtual Tracker: Device and  to_device()

  #### 1. The Problem

  Because MLX arrays reside in unified memory, they do not belong to a device and do not have  .device  attributes or  .to_device()  methods.
  However, the Array API requires  device(to_device(x, dev)) == dev  to hold true. If  to_device  just returned the array (since nothing physically
  moves), any subsequent call to check the device would simply return the active system default (usually  gpu ), causing device tests to fail when CPU is
  selected.

  #### 2. The Solution

  We implemented a virtual device registry using an  id(array) -> Device  map in Python:

    _mlx_device_map = {}

    def _mlx_cleanup(ref_id):
        _mlx_device_map.pop(ref_id, None)

    def _set_mlx_device(x, device):
        ref_id = id(x)
        _mlx_device_map[ref_id] = device
        # Prevent memory leaks: clean up when the array is garbage collected
        weakref.finalize(x, _mlx_cleanup, ref_id)

  • No  __eq__  collisions: We use the integer  id(x)  rather than a standard  WeakKeyDictionary  because looking up an array in a  WeakKeyDictionary
  invokes  x == y  on keys. Since MLX comparisons return element-wise boolean arrays rather than a single Python boolean, dictionary lookups would raise a
  ValueError .
  • No memory leaks: Using  weakref.finalize  automatically pops the ID from the tracking dictionary when the array is garbage collected.		
  
  Note: Some help of AI was used for code and pr description writing :)

### DIFF
--- a/.github/workflows/mlx-tests.yml
+++ b/.github/workflows/mlx-tests.yml
@@ -1,0 +1,24 @@
+name: MLX Tests
+
+on: [push, pull_request]
+
+jobs:
+  mlx-tests:
+    runs-on: macos-14 # Apple Silicon runner
+    steps:
+      - uses: actions/checkout@v7.0.0
+
+      - name: Set up Python
+        uses: actions/setup-python@v6.2.0
+        with:
+          python-version: '3.13'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest mlx numpy array-api-strict ndonnx
+
+      - name: Run MLX Compat Tests
+        run: |
+          export PYTHONPATH="${GITHUB_WORKSPACE}/src"
+          pytest tests/test_mlx.py tests/test_common.py tests/test_isdtype.py tests/test_copies_or_views.py -k mlx -v
--- a/meson.build
+++ b/meson.build
@@ -61,6 +61,15 @@ sources_raw = {
     'src/array_api_compat/torch/fft.py',
     'src/array_api_compat/torch/linalg.py',
   ],
+
+  'array_api_compat/mlx': [
+    'src/array_api_compat/mlx/__init__.py',
+    'src/array_api_compat/mlx/_aliases.py',
+    'src/array_api_compat/mlx/_info.py',
+    'src/array_api_compat/mlx/_typing.py',
+    'src/array_api_compat/mlx/fft.py',
+    'src/array_api_compat/mlx/linalg.py',
+  ],
 }
 
 sources = {}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,6 +55,7 @@ dev = [
     "pytest",
     "torch",
     "sparse>=0.15.1",
+    "mlx"
 ]
 
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -55,7 +55,7 @@ dev = [
     "pytest",
     "torch",
     "sparse>=0.15.1",
-    "mlx"
+     "mlx; platform_system == 'Darwin' and platform_machine == 'arm64'"
 ]
 
 
--- a/src/array_api_compat/common/_helpers.py
+++ b/src/array_api_compat/common/_helpers.py
@@ -273,6 +273,29 @@ def is_pydata_sparse_array(x: object) -> TypeIs[sparse.SparseArray]:
     return _issubclass_fast(cls, "sparse", "SparseArray")
 
 
+def is_mlx_array(x: object) -> bool:
+    """
+    Return True if `x` is an MLX array.
+
+    This function does not import MLX if it has not already been imported
+    and is therefore cheap to use.
+
+    See Also
+    --------
+
+    array_namespace
+    is_array_api_obj
+    is_numpy_array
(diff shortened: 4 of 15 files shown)


ev-br:
Thanks for the PR. 

We'll need some more context though. I see that it's been discussed for a while, https://github.com/data-apis/array-api-compat/issues/162, and a suggested avenue was that MLX adds Array API support in its main namespace, similar to JAX.
And there are several discussions in MLX itself,  https://github.com/ml-explore/mlx/issues/3514 and references therein.
So what is the upstream status? Upstream adding support is of course the ideal outcome.

Next, what is the current compatibility level of MLX and Array API? I ran the test suite against this PR (`$ ARRAY_API_TESTS_MODULE=array_api_compat.mlx pytest array_api_tests/test_creation_functions.py  -v`), and it fails all around. 
Some of these failures might be deficiencies in the test suite itself (cf https://github.com/data-apis/array-api-tests/issues/431#issuecomment-4183743987). 

I'd suggest to start from the beginning:
- create an overview document with a list of Array API functions/objects and their MLX analogs (e.g., three columns: an Array API function, a matching MLX function, rough compat status---fundamental issues if any, otherwise name/signature mismatches, arguments, defaults and so on).
- For matching functions, run the corresponding tests from the test suite. Does it pass? If not, what are issues, is the test suite at fault or is MLX incompatible?
- For functions with small differences, would MLX itself accept changes or aliases? (concatenate/concat, acos/arccos etc)
- For situations where the problem is with the test suite, what is missing in it?
- For fundamental problems, what would it take to fix (e.g., monkey-patching the array object we 99/100 won't do here, but let's make it explicit in the design overview)

Overall, I'm pretty sure that the support is doable (at a level comparable with e.g. jax which implements a part of the standard, and pytorch which also implements a part of the standard and has some of the inconsistencies smoothened out in the compat layer). Making it work will take a bit of effort though, spread between the MLX itself, the Array API workgroup if needs be, the test suite and maybe indeed the compat layer. But it all starts with an explicit context and a design overview.

aaishwarymishra:
My bad, I kind of discussed this with Ralf but you are right, I should get to be more organised. I think most of the methods are implemented now in mlx, accept few and some differences in how the implemented methods behave.

Well the biggest problem right now is no .device() support, and the no float64 support on GPU.

Well float64 can be added in the future but I think we might have to implement the device work around .

Also I will make a nice detailed doc so it would be clearer :)

rgommers:
A detailed doc sounds quite useful.

> Well the biggest problem right now is no .device() support, and the no float64 support on GPU.

There are other libraries with no `.device` method, which is why SciPy uses the `xp_device` function as a helper - I don't think this is a major issue. 

No `float64` on GPU also isn't unique, and will be allowed in the next version of the standard, see the just-merged https://github.com/data-apis/array-api/pull/1005.

rgommers:
MLX has done a lot of work, my understanding is that all the critical gaps were plugged, and if there are some missing routines, they're open to adding them.

Issue #452, opened 2026-07-18
MLX compatibility: Creation functions


aaishwarymishra:
| Array API | MLX Analog | Status | Blame / Notes | Upstream | Test Node | Result |
| --- | --- | --- | --- | --- | --- | --- |
| `arange(start, /, stop=None, step=1, *, dtype=None, device=None)` | `arange(start, stop, step, dtype=None, *, stream=None)` | partial | `stop=None` now accepted. Remaining failure: integer args larger than signed 32-bit (e.g. `step=2**31`) — the nanobind binding uses C++ `int`, so `xp.arange(n-1, n+3)` with `n = iinfo(int32).max` raises `TypeError`. Still no `device`. | ml-explore/mlx#3982 (merged); int32 limit fixed ml-explore/mlx#4255| `test_arange` | Passed |
| `asarray(obj, /, *, dtype=None, device=None, copy=None)` | `asarray(a, dtype=None, *, copy=None)` | partial — won't fix | `copy=False` raises even for an MLX array already of the requested dtype. Fix was proposed and closed unmerged. No `device`; `dtype` not keyword-only. | ml-explore/mlx#4036 (closed, not merged) | `test_asarray_arrays` | Failed |
| `empty(shape, *, dtype=None, device=None)` | `empty(shape, dtype=float32, *, stream=None)` | undocumented | Implemented (pure alias) but still absent from MLX docs. No `device`. | ml-explore/mlx#3729 (merged) | `test_empty` | Passed |
| `empty_like(x, /, *, dtype=None, device=None)` | `empty_like(a, /, *, dtype=None, stream=None)` | undocumented | `dtype` now accepted — `empty_like` aliases `zeros_like`, so it inherited the fix. Still undocumented; no `device`. | ml-explore/mlx#4028 (merged); #4037 (closed, superseded) | `test_empty_like` | Passed |
| `eye(n_rows, n_cols=None, /, *, k=0, dtype=None, device=None)` | `eye(n, m=None, k=0, dtype=float32, *, stream=None)` | partial | `eye(0)` fixed upstream; the too-large-`k` failure was a test-suite issue, fixed at the tip of array-api-tests. Remaining failure: GPU scatter does not support `uint64` MLX limitation, so identity creation fails for that dtype. | ml-explore/mlx#3952 (merged); uint64 case cant be supported due to mlx limitation ml-explore/mlx#4300 | `test_eye` | Failed |
[...]

PR #453, opened 2026-07-27
Add initial MLX support with the creation wrappers

aaishwarymishra:
This pull request introduces initial support for the MLX backend in the `array_api_compat` package. The main changes add MLX-specific modules, implement compatibility wrappers for core array functions, and update the build and development configuration to include MLX. Below are the most important changes:

For #452 

**MLX Backend Implementation:**

* Added the new `array_api_compat/mlx` backend with `__init__.py`, `_aliases.py`, and `_typing.py` to support MLX arrays and provide compatibility wrappers for common array operations.
* Implemented MLX-compatible versions of `arange`, `asarray`, `empty_like`, `eye`, `meshgrid`, `ones_like`, and `zeros_like` in `_aliases.py`, handling MLX-specific quirks and edge cases.
* Defined MLX-specific type aliases (`Array`, `DType`, `Device`) in `_typing.py` for consistent type handling.
* Updated `__init__.py` to expose all relevant symbols and ensure proper module structure for the MLX backend.

**Build and Development Environment:**

* Updated `meson.build` and `pyproject.toml` to include MLX sources and conditionally require the `mlx` package on macOS (`sys_platform == 'darwin'`). [[1]](diffhunk://#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4R64-R68) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R58)

### DIFF
--- a/src/array_api_compat/mlx/_aliases.py
+++ b/src/array_api_compat/mlx/_aliases.py
@@ -41,16 +41,18 @@ def arange(
 ):
     """
     Note on MLX Compatibility:
-        Native `mx.arange` fails with Python `int` arguments exceeding 2**31 - 1
-        due to underlying C++ type binding limits. Large bounds must be cast or
-        handled directly by this wrapper.
+        Native `mx.arange` rejects some Python `int` arguments above 2**31 - 1.
+        This wrapper handles empty and single-element ranges directly; larger
+        int64 and uint64 ranges remain unsupported by MLX.
     """
     if stop is None:
         stop = start
         start = 0 if isinstance(stop, int) else 0.0
     # MLX defaults to float32 even when all arguments are integers,
     # so we must explicitly set the dtype to int32 in that case.
-    is_float = isinstance(start, float) or isinstance(stop, float) or isinstance(step, float)
+    is_float = (
+        isinstance(start, float) or isinstance(stop, float) or isinstance(step, float)
+    )
     dtype = dtype if dtype is not None else (mx.float32 if is_float else mx.int32)
     # MLX may overflow while converting a large step before determining that the
     # result has at most one element (for example, arange(0, 1, 2**31) is [0]),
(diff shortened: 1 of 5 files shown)


aaishwarymishra:
After adding temporary compatibility wrappers and rerunning `test_creation_functions.py`, I found a few additional MLX limitations:

* **`arange` dtype:** MLX defaults integer inputs to `float32`, while the Array API expects `int32`. The wrapper uses `int32` for integer-only inputs and `float32` otherwise.
* **Large integer `arange`:** `mx.arange` rejects some values above `2**31 - 1`. Empty and single-element ranges can be handled, but nontrivial `int64` and `uint64` ranges still fail.
* **`eye` with `int64`/`uint64`:** MLX GPU scatter does not support these dtypes, so the wrapper creates the matrix in `int32` and casts it.
* **`asarray(copy=False)`:** Returning an existing compatible MLX array avoids a copy error, but the aliasing test fails because MLX does not support 0-D assignment with `x[...] = value`.
* **`empty_like`, `ones_like`, `zeros_like`:** Wrappers use `x.dtype` when `dtype=None`.
* **`meshgrid`:** The wrapper converts MLX’s list output to the required tuple.
* **`linspace`:** Still fails because MLX does not support `endpoint`. No wrapper was added due to possible numerical differences.

With:

```sh
ARRAY_API_TESTS_SKIP_DTYPES=float64,complex64,int64,uint64
```

the suite reports **14 passed, 2 failed**:
*Note: Some times we got lucky with arange and other methods due to hypothesis gives different values which sometimes passes the tests*

1. `test_asarray_arrays`: unsupported 0-D `Ellipsis` assignment.
2. `test_linspace`: missing `endpoint` support.


*NOTE:  MLX does not implement __eq__ for comparing None with datatype, also we skipped complex completely due to problem with `complex()` but tests doesn't allow completely skipping complex, I think tests can be improved here*

ev-br:
Thanks!

Going forward, it would be helpful to list reproducible examples from the test suite failures, which the test suite prints on a failure. 

> complex() but tests doesn't allow completely skipping complex, I think tests can be improved here

Please open an issue in the test suite, with an MRE. If it's a test suite problem, it should be reproducible with numpy or array-api-strict, I suppose.

> MLX does not implement eq for comparing None with datatype

This is one of things to propose adding to MLX; even if they decline, we'll have an answer. 
Similar for other fixes, large and small.

src/array_api_compat/mlx/_aliases.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def ones_like(
+    x: Array,
+    /,
+    *,
+    dtype: None | DType = None,
+    device: None | Device = None,
+):
+    return mx.ones(x.shape, dtype=x.dtype if dtype is None else dtype, stream=device)
```
ev-br, line 111: `stream=device` is a clever trick indeed.

At this stage I'm not sure what the consequences are, and whether we should do it across the board, or instead keep accepting the `stream` argument as a fall-through via `**kwargs`, as done for other arguments beyond the array API spec.

ev-br (commented):
Given that the upstream work in MLX is proceeding at a high pace, the majority of workarounds here are no longer needed.
One other feature of this PR, the `device=stream` trick needs more analysis: it's a clever trick which might turn out to be bad idea in the end.

All in all, I think it's best to park this PR and concentrate on improving the compatibility upstream. 

Of course, if this sort of WIP PR is useful to you as a testbed for your MLX work, feel free to keep using it @aaishwarymishra. If you do this, it'd be best to keep it current and remove things which are already upstream. This way, it would be easier for everybody to assess the compatibility status.

aaishwarymishra:
Yeah i will make it draft for now :)

PR #454, opened 2026-07-27
ENH: use native numpy.reshape for numpy>=2.1

stanbot8:
`reshape` with `copy=False` assigns to the array shape attribute that numpy 2.5 deprecates.

Bind the numpy namespace `reshape` to the native numpy function from numpy 2.1. That release added the copy argument. Keep the shared workaround for an older numpy.

Leave the cupy and dask paths unchanged. Neither namespace accepts a copy argument, so both still need the shared workaround.

### DIFF
--- a/src/array_api_compat/numpy/_aliases.py
+++ b/src/array_api_compat/numpy/_aliases.py
@@ -50,7 +50,10 @@ var = get_xp(np)(_aliases.var)
 cumulative_sum = get_xp(np)(_aliases.cumulative_sum)
 cumulative_prod = get_xp(np)(_aliases.cumulative_prod)
 permute_dims = get_xp(np)(_aliases.permute_dims)
-reshape = get_xp(np)(_aliases.reshape)
+if np.__version__ < '2.1':
+    reshape = get_xp(np)(_aliases.reshape)
+else:
+    reshape = np.reshape
 argsort = get_xp(np)(_aliases.argsort)
 sort = get_xp(np)(_aliases.sort)
 nonzero = get_xp(np)(_aliases.nonzero)
--- a/tests/test_cupy.py
+++ b/tests/test_cupy.py
@@ -1,5 +1,6 @@
 import pytest
 from array_api_compat import device, to_device
+import warnings
 
 xp = pytest.importorskip("array_api_compat.cupy")
 from cupy.cuda import Stream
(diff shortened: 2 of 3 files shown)


ev-br:
Development in this repository is human-first,  and we very much prefer to not receive automated contributions from bots.

stanbot8:
@ev-br I am a human contributor. `stanbot8` is only my username. I reviewed and tested this change and can explain or revise each part of it.

ev-br:
Than my apologies @stanbot8 . We receive too many automated pull requests and the default shifts to "shoot first ask later". And the combination of the GH username and the PR style was too LLM-like.

So, could you then please explain (preferably in your human words, without using an LLM), what is this patch, what is the problem it aims to address and what is the solution?

stanbot8:
No worries @ev-br, I understand how my username and the PR style could have given that impression.

I found this while testing the project with NumPy 2.5.1. The `copy=False` path directly assigned to `y.shape`, which NumPy 2.5 deprecated. My patch uses the namespace's `reshape(..., copy=False)` when it supports the `copy` argument and keeps the existing fallback otherwise.

ev-br:
Oh I see what's going on here now, thank you!

So the `copy=` behavior of  `reshape` function is effectively untested (https://github.com/data-apis/array-api-tests/pull/432) and subtly broken in several respects. For one, the `copy=False` workaround drops `**kwargs`.

Now, using the native function instead of a pile of workarounds is definitely a win and is very worthy goal to work towards!

I just quickly tried running tests with bare numpy 2.4.2 and a branch of https://github.com/data-apis/array-api-tests/pull/432/ and it passes:

```
$ ARRAY_API_TESTS_MODULE=numpy pytest array_api_tests/test_manipulation_functions.py::test_reshape -v --max-examples=10_000
```

Therefore, how about:

- branch on a numpy version at https://github.com/data-apis/array-api-compat/blob/main/src/array_api_compat/numpy/_aliases.py#L53 : only use the existing workaround for older numpies and use the native reshape for newer numpies; this would remove the need for `inspect`-ing things at runtume;
- for CuPy and `dask`, keep the existing workaround, too (ain't broken and all) --- for CuPy we'll be able to simplify in a year or so:  https://github.com/cupy/cupy/pull/10109

For testing, I *think* it would be cleaner to test numpy and cupy instead of `common`, a test with `shares_memory` would be great.

If you are feeling brave and have cycles to spare, help fixing https://github.com/data-apis/array-api-tests/pull/432 would be appreciated but is certainly not a blocker for this PR. 
If you are feeling *really* brave, `torch/_aliases.py::reshape` simply raises with `copy=`, but that'd even more of a "not a blocker".

stanbot8:
Rewrote per your suggestion.

The numpy version branch sends only an older numpy through the workaround, so the runtime `inspect` is gone. I moved the tests out of `common` into the numpy and cupy files, and both assert `shares_memory`. My machine has a CUDA GPU, so the cupy test runs instead of skipping.

ev-br:
CI errors seem unrelated. https://github.com/data-apis/array-api-compat/pull/456 is to fix the error with `-strict`, not sure yet what's going on with jax and its Array API version.

ev-br:
close/reopen to retrigger CI after gh-456

ev-br (approved):
LGTM now, thank you @stanbot8 ! Merging.

PR #457, opened 2026-07-28
ENH: torch.reshape: implement copy with Torch view and clone operations

stanbot8:
The Torch `reshape` wrapper rejects each explicit copy value.

Implement `copy` with Torch `view` and `clone` operations.

[array-api-tests#432](https://github.com/data-apis/array-api-tests/pull/432) passed with 10000 generated examples.

All tests pass.


ev-br (approved):
LGTM!
Thank you @stanbot8 , merging.

PR #459, opened 2026-07-29
fix: define dask reshape copy behavior

stanbot8:
The shared Dask `reshape` wrapper fails for scalar and 0-size `copy=False` cases.

This change routes Dask `reshape(..., copy=False)` to the native lazy operation and enables the Array API `reshape` test.

This change fixes the Dask issues at [issue 446](https://github.com/data-apis/array-api-tests/issues/446).


ev-br:
This currently swallows copy=False, no?

stanbot8:
Yes, it currently treats `copy=False` like `copy=None`.

I've opened a [pull request to Dask](https://github.com/dask/dask/pull/12529) that adds the `copy` contract.

I will keep this PR as a draft and update the wrapper to forward `copy` if Dask accepts the API.

PR #468, opened 2026-08-23
PERF: fft: unnecessary copy in `.astype` if the output array is already the…

Nin17:
… correct dtype.

The `fft` functions unnecessarily copy the return array if the `dtype` is the desired `dtype`.
I believe the use of `.astype` is to enable support for numpy < 2.0. As support for `"f4"` and `"c8"` dtypes was introduced in numpy 2.0 the unnecessary copies only occur in numpy >= 2.0.
I'm not sure when cupy stopped upcasting `"f4"` and `"c8"` dtypes in ffts, but it currently doesn't, and therefore also has the unnecessary copy.

### DIFF
--- a/src/array_api_compat/common/_fft.py
+++ b/src/array_api_compat/common/_fft.py
@@ -21,7 +21,7 @@ def fft(
 ) -> Array:
     res = xp.fft.fft(x, n=n, axis=axis, norm=norm)
     if x.dtype in [xp.float32, xp.complex64]:
-        return res.astype(xp.complex64)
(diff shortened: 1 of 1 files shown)


Nin17:
The numpy test fails:
`
FAILED array_api_tests/test_linalg.py::test_trace
`
Seems unrelated to these changes.

ev-br:
CI failures are unrelated, https://github.com/data-apis/array-api-tests/issues/455

ev-br (approved):
LGTM thank you @Nin17

PR #472, opened 2026-08-27
Remove caching from torch information endpoint

### LINKED ISSUE #470: PyTorch namespace info caching doesn't work
In `src/array_api_compat/torch/_info.py` some methods on the `__array_namespace_info__` class have a `@cache`. The other namespaces don't have that and as far as I can tell the caching doesn't do what it should :-/ Should we remove it?

```python
In [1]: import array_api_compat.torch as xp

In [2]: Info = xp.__array_namespace_info__

In [3]: Info().dtypes(kind="real floating")
Out[3]: {'float32': torch.float32, 'float64': torch.float64}

In [4]: Info().dtypes(kind="real floating")
Out[4]: {'float32': torch.float32, 'float64': torch.float64}

In [5]: Info().dtypes(kind="real floating")
Out[5]: {'float32': torch.float32, 'float64': torch.float64}

In [6]: Info.dtypes.cache_info()
Out[6]: CacheInfo(hits=0, misses=3, maxsize=None, currsize=3)
```
I was expecting to see some `hits` with this if the cache was working.

We could turn `__array_namespace_info__` into a singleton and add a layer of indirection to `dtypes` so that every caller still gets their own instance of the `dict`. This avoids everyone sharing the same instance. For `devices` we don't need that because it returns a `tuple`.

Something like this:

```diff
--- a/array_api_compat/torch/_info.py
+++ b/array_api_compat/torch/_info.py
@@ class __array_namespace_info__:
     __module__ = 'torch'
 
+    _instance = None
+
+    def __new__(cls):
+        # `devices` and `_dtypes_for` are memoized with `functools.cache`,
+        # which keys on `self`. Using a singleton means the cache is used
+        # for repeated calls.
+        if cls._instance is None:
+            cls._instance = super().__new__(cls)
+        return cls._instance
+
     def capabilities(self):
@@ def dtypes(self, *, device=None, kind=None):
-    @cache
     def dtypes(self, *, device=None, kind=None):
         """
         The array API data types supported by PyTorch.
 
         ... (docstring unchanged) ...
 
         """
+        # The copy is deliberately outside the memoized call. The cached dict
+        # is shared by every caller, and a caller that mutates what it was
+        # handed must not corrupt the next one.
+        return dict(self._dtypes_for(device=device, kind=kind))
+
+    @cache
+    def _dtypes_for(self, *, device=None, kind=None):
         res = self._dtypes(kind)
         for k, v in res.copy().items():
             try:
                 torch.empty((0,), dtype=v, device=device)
             except:
                 del res[k]
         return res

```

betatim:
The `@cache` decorator was not working so we remove it. Because it uses `self` as part of they key each new instance had a new fresh cache. And basically you get a new instance every time.

Closes #470

### DIFF
--- a/src/array_api_compat/torch/_info.py
+++ b/src/array_api_compat/torch/_info.py
@@ -9,7 +9,6 @@ more details.
 """
 import torch
 
-from functools import cache
 
 class __array_namespace_info__:
     """
@@ -249,7 +248,6 @@ class __array_namespace_info__:
             return res
         raise ValueError(f"unsupported kind: {kind!r}")
 
-    @cache
     def dtypes(self, *, device=None, kind=None):
         """
         The array API data types supported by PyTorch.
@@ -313,7 +311,6 @@ class __array_namespace_info__:
                 del res[k]
         return res
 
-    @cache
     def devices(self):
         """
         The devices supported by PyTorch.


ev-br:
LGTM and CI is green and it's merged. Thanks @betatim

betatim:
Out of interest: where/how is it configured that copilot comes and reviews? And can you configure copilot to tell it things about how to review?

ev-br:
Yeah, we should prbably just disable these automated Copilot revews
Based on https://github.com/orgs/community/discussions/189857, there's a hierarchy of org level > repo level > user level. For the first two, we need to ask the admins. @kgryte or @rgommers -- could you please disable Copilot reviews for the data-apis org and the array-api-compat repository?

rgommers:
Disabled at the repo level. There is no org-level setting for Copilot it looks like.

betatim:
While this review was not that insightful, but then the diff was pretty boring as well, on some projects I work on we have CodeRabbit which needed quite a bit of configuring (to make it good) and now does seem to come up with some useful information.

So I am a bit curious to see these automatic reviews, how to fine tune them and see what happens. What I am wondering is if they can be like some kind of "super linter" - they point things out that the contributor can address themselves so that when a human maintainer arrives the basics are taken care of.

ev-br:
>... CodeRabbit which needed quite a bit of configuring (to make it good) and now does seem to come up with some useful information.

Oh interesting! Care to share some links, here or elsewhere?
My experience with greptile-app, CodeRabbit, Copilot is that of low quality, low signal/noise ratio, and frankly annoying back-and-forth. But I never tried to fine-tune them, so I'm curious!

betatim:
in the cuml repo is where I interact with it most

https://github.com/NVIDIA/cuml/blob/main/.coderabbit.yaml is the config file we use, taken from cuopt and then tweaked - I think. For the interaction/usefulness part I think you'll have to click yourself through some of the PRs (no better way)

